PR #13, opened 2014-12-03
Use the new build env on Travis

joshk:
more cpu and ram, better networking, faster boot times

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,18 @@
 language: erlang
-services:
-  - mysql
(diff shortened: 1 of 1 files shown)


joshk:
FYI: I believe the new images have been deployed with the newer Erlang version.

zuiderkwast:
Ah, ok, so just `sudo: false` will enable this? I'll do it some time.

joshk:
Confirm :)


—

PR #17, opened 2015-04-01
add compartibility with mysql versions like 5.5.33a

umatomba:
With versions like 5.5.33a-MariaDB i have connection problem:

```
** exception exit: badarg
     in function  binary_to_integer/1
        called as binary_to_integer(<<"33a">>)
```


zuiderkwast:
Merged. Thanks!

PR #26, opened 2015-08-19
Support for strings as prepared statement parameters

SergejJurecko:


src/mysql_protocol.erl
```diff
@@ -736,6 +736,8 @@ encode_param(null) ->
 encode_param(Value) when is_binary(Value) ->
     EncLength = lenenc_int_encode(byte_size(Value)),
     {<<?TYPE_VAR_STRING, 0>>, <<EncLength/binary, Value/binary>>};
+encode_param(Value) when is_list(Value) ->
+    encode_param(unicode:characters_to_binary(Value));
```
zuiderkwast: Thanks for this PR. At first it looked like a simple merge to accept but after looking into it more i have a few concerns...

SQL queries as the query parameter to `mysql:query/2,3,4` are documented as being iodata as opposed to unicode character data.

In `mysql:encode/2`, lists are interpreted as iolists: https://github.com/SergejJurecko/mysql-otp/blob/master/src/mysql_encode.erl#L20 as opposed to unicode character data. I think it is important that the behaviour is identical whether prepared statements are used or encode/2 is used. The return value of mysql:encode/2 is also an iolist that can be used to build queries using simple iolist construction such as `["UPDATE foo SET bar = ", mysql:encode(Conn, Bar), " WHERE id = 1;"]`.

Probably it would make sense to changes all these unicode character data but that would break current behaviour and would mean that we have to do a major version bump if we strictly are to follow semver. Is it worth it? Can we do it with a just minor version bump?

Also (assuming the iodata/unicode issue can be resolved) I'd like a testcase (or a extra line in an existing testcase) to cover `encode_param(Value) when is_list(Value)`.

SergejJurecko:
Would it break current behavior? Unicode has the same numbers as ascii for ascii characters. Sending unicode to mysql:encode or mysql:execute results in a crash atm. I would consider replacing  iolist_to_binary with unicode:characters_to_binary as a bugfix. Though I have been known to be too optimistic :)

zuiderkwast:
After sleeping on this, I now remember why we can't treat everything as unicode: binary data! When sending binary blob data to the server, this is done as a string which is not valid unicode string so unicode:characters_to_binary would crash on this.

You're right about ascii strings. For these nothing breaks, but things would break for non-ascii utf-8 encoded strings.

zuiderkwast:
I'm still positive about this PR. We can require blob data to be sent as binaries and only do unicode:characters_to_binary for lists. SQL queries must still be iodata as they can contain non-unicode blob data in SQL string literals.

This is what I'd like before merging:

* unicode conversion for lists in mysql_encode:encode, so that the behaviour is the same as for prep statements;
* test for the added lines, preferably with some non-ascii characters;
* documentation for this behaviour in "value representation" section on the edoc overview page.

For the documentation point, I'd like to change the type for "VARCHAR, TEXT, etc" from iodata() to unicode:chardata() and add a row for "VARBINARY, BLOB, etc" under the "VARCHAR, TEXT, etc" row with type binary(). The footnote 1 should be edited also to describe that unicode conversion is done for lists but not for binaries.

SergejJurecko:
OK. I will get on it sometime this week.

SergejJurecko:
Hmm. Not sure what to do about the R16 thing. I guess unicode module produced something different back then.

SergejJurecko:
Oh right. Forgot about that. Added the line.

zuiderkwast:
Thanks! I'll edit the docs a little. It doesn't mention the type `unicode:chardata()`, etc.

PR #34, opened 2015-12-03
Link processes that are doing transactions

KoFish:
In the case of pool handling and supervisors a connection can be placed in a
invalid state if a process is using a checked out connection and then is
brutally killed by, for example, its supervisor. To avoid this a link between
the calling process and the connection is being kept while performing the
transaction so that the connection process is killed in case the calling process
dies.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -373,11 +373,11 @@ transaction(Conn, Fun, Args, Retries) when is_list(Args),
     %% The guard makes sure that we can apply Fun to Args. Any error we catch
     %% in the try-catch are actual errors that occurred in Fun.
     ok = gen_server:call(Conn, start_transaction),
-    try apply(Fun, Args) of
-        ResultOfFun ->
-            %% We must be able to rollback. Otherwise let's crash.
-            ok = gen_server:call(Conn, commit),
-            {atomic, ResultOfFun}
+    Result = try apply(Fun, Args) of
+                ResultOfFun ->
+                    %% We must be able to rollback. Otherwise let's crash.
+                    ok = gen_server:call(Conn, commit),
+                    {atomic, ResultOfFun}
     catch
         throw:{implicit_rollback, N, Reason} when N >= 1 ->
             %% Jump out of N nested transactions to restart the outer-most one.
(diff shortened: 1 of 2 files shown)


src/mysql.erl
```diff
@@ -422,7 +422,9 @@ transaction(Conn, Fun, Args, Retries) when is_list(Args),
                 exit  -> Reason
             end,
             {aborted, Aborted}
-    end.
```
zuiderkwast: How about exceptions raised in the `catch` clause? I think `gen_server:call(Conn, end_transaction)` is better called in an `after` clause, i.e. `try ... catch ... after gen_server:call(Conn, end_transaction) end`.

src/mysql.erl
```diff
@@ -422,7 +422,9 @@ transaction(Conn, Fun, Args, Retries) when is_list(Args),
...
+    end,
+    ok = gen_server:call(Conn, end_transaction),
+    Result.
```
zuiderkwast: How about exceptions raised in the `catch` clause? I think `gen_server:call(Conn, end_transaction)` is better called in an `after` clause, i.e. `try ... catch ... after gen_server:call(Conn, end_transaction) end`.

src/mysql.erl
```diff
@@ -668,11 +670,12 @@ handle_call(backslash_escapes_enabled, _From, State = #state{status = S}) ->
     {reply, S band ?SERVER_STATUS_NO_BACKSLASH_ESCAPES == 0, State};
 handle_call(in_transaction, _From, State) ->
     {reply, State#state.status band ?SERVER_STATUS_IN_TRANS /= 0, State};
-handle_call(start_transaction, _From,
+handle_call(start_transaction, {From, _Tag},
             State = #state{socket = Socket, transaction_level = L,
                            status = Status})
   when Status band ?SERVER_STATUS_IN_TRANS == 0, L == 0;
        Status band ?SERVER_STATUS_IN_TRANS /= 0, L > 0 ->
+    erlang:link(From),
```
zuiderkwast: How about nested transactions? We only want to create a link if it is the outermost transaction, right? (Or is it possible to have multiple links between two processes?)

src/mysql.erl
```diff
@@ -681,6 +684,9 @@ handle_call(start_transaction, _From,
                                                ?cmd_timeout),
     State1 = update_state(Res, State),
     {reply, ok, State1#state{transaction_level = L + 1}};
+handle_call(end_transaction, {From, _Tag}, State) ->
+    erlang:unlink(From),
+    {reply, ok, State};
```
zuiderkwast: Nested transactions? (See comment for link in start_transaction.)

zuiderkwast:
I'm not sure I like the idea of make code safe for brutal kill. That seems too defensive to me somehow. How is this problem handled in other database drivers?

Besides, it is always possible to leave a connection in a weired state by changing some mysql session variables or by calling starting a transaction manually using query("BEGIN")...

PR #35, opened 2016-01-08
change required for elixir/exrm tool

deanzone:
This tag is required for the elixir/exrm tool which uses systools/relx for deployment packaging.


zuiderkwast:
Thanks!

Next time please use a more descriptive commit message. "This tag" is impossible to understand without seeing the code. The commit messages are used for the change log which is updated using `make CHANGELOG.md`. See https://github.com/mysql-otp/mysql-otp/blob/master/CHANGELOG.md

Issue #33, opened 2015-10-09
Detect server shutdown


zuiderkwast:
When the server shuts down cleanly it sends a TCP message to notify us that it is closing the connection. We do not detect this because our gen_tcp connection is in passive mode.

Solution: Set the connectionn to `{active, once}` and listen for info message `{tcp_closed, _Socket}`. Stop the gen_server when this happens (so that some supervisor or pool can start a new connection). Before any communication with the server (such as when we send a query) we should set the connection back to passive mode, do the communication and afterwards set it back to active once.

Note that we cannot detect a network split or a server crash. In this case there is no TCP message that tells us that the connection is gone.

Related issues: #22, #25 and https://github.com/mysql-otp/mysql-otp-poolboy/issues/14

sngyai:
Has it been solved?

zuiderkwast:
No. How important is this to you? A pull request is welcome. I would bump the version to 2.0.0 when this feature is solved.

newsdy:
Has it been solved? I have the same problem

alexshavelev:
think i will do it in 1-2 days

alexshavelev:
#39 hope it will help

zuiderkwast:
Closed by PR #41 (which was greatly inspired by #39). Thanks @johlo and @alexshavelev!

zuiderkwast:
I wrote earlier that I would bump the version to 2.0.0 when this feature is solved, but since this feature is not breaking any backwards compatibility, there is no reason to bump the major version. I have bumped the minor version to 1.2.0.

PR #39, opened 2016-05-17
fix issue #33

alexshavelev:


alexshavelev:
please look for tests https://travis-ci.org/mysql-otp/mysql-otp/jobs/130820684
fails dont depend on improvements

zuiderkwast:
Nice! Any clue why the tests fail at Travis? Is mysql is not running? I'd like to fix that before I merge this PR.

src/mysql.erl
```diff
@@ -479,12 +479,15 @@ init(Opts) ->
...
     %% Connect socket
-    SockOpts = [{active, false}, binary, {packet, raw} | TcpOpts],
+    SockOpts = [binary, {packet, raw} | TcpOpts],
     {ok, Socket} = gen_tcp:connect(Host, Port, SockOpts),
 
     %% Exchange handshake communication.
     Result = mysql_protocol:handshake(User, Password, Database, gen_tcp,
                                       Socket),
+    inet:setopts(Socket, [{active, once}]),
+    MonitorRef = erlang:monitor(process, self()),
```
zuiderkwast: Why monitor ourself? If we are dead, we won't get the 'DOWN' message anyway, do we?

zuiderkwast:
The failing tests are because of the changes in this branch. You forgot some places where the connection needs to be set to passive. The first one is in mysql_protocol:handshake (on the first recv) because the connection is in active mode here. We need to connect with {active, false} like before. The other fails are also because the connection needs to be set to passive before the calls to in mysql_protocol:quit and mysql_protocol:ping and set back to active once afterwards. Please try and make the tests working.

Also please explain why we are monitoring ourself. I don't understand this.

Thanks for your effort!

zuiderkwast:
Any news on this?

zuiderkwast:
Closing this in favor of #41. Thanks!

PR #41, opened 2016-06-11
Stop the gen_server when mysql server closes connection

johlo:
Another PR for handling issue #33, very similar to PR #39 but no new eunit failures are introduced.
Note that I have 7 failing unit tests, but the same tests fail on a pristine master so I assume it is unrelated to this patch. I'm testing on OS X 10.11 with Mysql 5.7.13.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -479,12 +479,14 @@ init(Opts) ->
     end,
 
     %% Connect socket
-    SockOpts = [{active, false}, binary, {packet, raw} | TcpOpts],
+    SockOpts = [binary, {packet, raw} | TcpOpts],
     {ok, Socket} = gen_tcp:connect(Host, Port, SockOpts),
 
     %% Exchange handshake communication.
+    inet:setopts(Socket, [{active, false}]),
     Result = mysql_protocol:handshake(User, Password, Database, gen_tcp,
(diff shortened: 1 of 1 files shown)


zuiderkwast:
This looks promising! It's simple and strait forward.

zuiderkwast:
Is there a reasonable way to test the code receiving `tcp_closed` and `tcp_error`? Would it make sense to let a test case send a message to the client process to fool it to terminate? I guess yes....

I have tried really hard to get coverage near 100% in this project. Wherever a line can not be covered by a test, I have described that in a comment, such as some cases that only occur for version 4.x.x of MySQL.

zuiderkwast:
The tests pass for me (MySQL 5.6.30 on Debian) and on Travis too, so no failing tests are introduced here. It would be interesting to know what failing tests you have though and why. Perhaps open a separate issue?

johlo:
@zuiderkwast It should be possible to add a eunit test that sets a very short timeout (e.g. `SET SESSION wait_timeout = 1`) and checks that we get the tcp_closed.
Not sure how to provoke a tcp_error though.

zuiderkwast:
Thanks for the suggestion! I have added a test using `SET SESSION wait_timeout = 1`.

Issue #31, opened 2015-09-11
replace error(old_server_version)


ingwinlu:
so I recently had the pleasure to deploy to an older mysql version(5.1) and encountered the old_server_version error.

It would be much nicer to try and continue if possible and just not offer some of the features that are tested during the handshake (I only needed 1 simple query to get some data).

If that is not reasonable, it would at least be good to know which versions of popular sql servers are supported and what features are missing for mysql-otp to work in the error output.

zuiderkwast:
5.1 really should work. If it doesn't this is definitely a bug. Any version from 4.1 and up should work.

ingwinlu:
since I wasn't the one deploying and only got the error log in a mail I can't really provide more input.
The person responsible said that they had `Server version: 5.1.56 MySQL Community Server (GPL) by Remi` installed (on centos 7).

Maybe it was one of the other requirements that lead to the error (i.e. no multi statements)? Which would reinforce my need to error out with the exact reason instead of just `old_server_version`.

zuiderkwast:
Right, it's the MULTI_* capability flags that caused the bug. It is actually caused by this bug in the server: http://bugs.mysql.com/bug.php?id=42268 . There is a workaround as you can read in the comments. I'm on it now.

zuiderkwast:
Too bad Travis can't run tests on multiple MySQL server versions. That would have been really useful.

Thanks for the report!

ingwinlu:
cheers, I will try and ask my client to deploy an updated version to see if that fixed it. They found a newer mysql version on another machine somewhere and moved the db there so I am not sure if they are still interested though.

ingwinlu:
Just got confirmation that the fix indeed resolved the issue we had in deployment. Thanks!

PR #42, opened 2016-06-14
Fix issue #33: mysql_socket to detect TCP Closed by Server

emqplus:
First of all, thanks @zuiderkwast for the awesome Erlang MySQL driver. I am the author of http://github.com/emqtt/. We received lots of issue reports after we use mysql-otp as MySQL driver in 0.17.x release:

https://github.com/emqtt/emqttd/issues/586

https://github.com/emqtt/emqttd/issues/523

https://github.com/emqtt/emqttd_plugin_mysql/issues/15

The root cause is mysql-otp use gen_tcp:recv to receive data passively. The client cannot receive tcp_closed event when MySQL server tries to shutdown idle clients.

We refactored the socket layer, which use '{active, once}' to receive tcp data like emqttc did:

Socket


emqplus:
@newsdy,  @alexshavelev, please help to verify the PR

RaoH:
Sorry of the absence, usual excuse of work is taking a lot of time!

I will test this MR locally and get back to you! But it looks good from my viewpoint =)

zuiderkwast:
Thanks for your effort @emqplus! It is a lot of changes. Refactoring makes it hard to see what has actually been changed and what has just been moved to other files. Apart from receiving messages in a receive loop using {active, once}, you have also implemented SSL and some other things that I don't know why they are done. When all comes in one PR it makes it hard to merge while keeping up the minimalist and clear approach of this project. Also there are some things that I don't understand such as why a separate controlling process is assigned and why the tcp_closed and other messages are received using a recursive receive loop (and why that process is hibernating) instead of in handle_info in the gen_server. To me, this makes the code harder to understand.

Due to the large changes I find it easier to accept a simple PR such as #41. Then I would be interested to see a separate PR for SSL.

emqplus:
Hi @zuiderkwast, 

1. A separate process which is responsible to receive and parse raw TCP data into protocol packets seems a design pattern for erlang socket layer. The pattern is wildly used and proved by some projects:

RabbitMQ: https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_reader.erl

ejabberd: https://github.com/processone/exmpp/blob/master/src/network/exmpp_socket.erl

emqttc: https://github.com/emqtt/emqttc

2. If we use the main mysql client process to receive TCP data, we have to call 'inet:setopts/2' to activate/deactivate the socket in many places.

3. The receiver process will hibernates to GC the refc binaries allocated for TCP data.

I'd like to remove the SSL supports from this PR and submit a new one. I hope we could invite more contributors to discuss and improve the socket layer for the issue #33 makes this driver cannot be used wildly by erlang communities.

Thanks for your awesome work. mysql-otp is the best Erlang MySQL driver, I think:)

zuiderkwast:
Thanks for your clarifications. There are only 14 places we have to call `inet:setopts/2` (twice for each of these places). As you can see in #41 there is not very much code required to do this. Considering we are only working with one connection I think the extra receiver process which needs to be garbage collected is adding complexity and I cannot see how this would be improving performance. Feel free to explain to me if I'm wrong.

emqplus:
@zuiderkwast, If the issue #31 could be fixed by PR #41, this PR is not necessary to merge. Usually the  design pattern is for asynchronous client driver, in which case the main client process will not block to wait for a response. I will upgrade the driver we used to 1.2.0 first, thanks.

zuiderkwast:
OK, great. Yes, I agree, for an asynchronous driver the socket layer would make more sense. I'm closing this in favour of #41 then.

PR #44, opened 2016-06-27
fix typo gitbub -> github

mdg:


zuiderkwast:
Thanks for finding this! The HTML pages in the gh-pages branch are generated. The typo should be fixed in doc/overview.edoc instead. I just did that myself. But thanks!

Issue #43, opened 2016-06-20
Several unit tests failing


johlo:
@zuiderkwast As I mentioned in my previous PR several unit tests are failing for me.
I'm running MySQL 5.7.13 (installed via brew) on Mac OS X 10.11.4.
Here's the full output from `make tests` on a pristine `master` checkout:

```
$ make tests
 APP    mysql.app.src
 GEN    test-dir
 GEN    eunit
mysql_tests: query_test_ (Text protocol)...*failed*
in function mysql_tests:text_protocol/1 (test/mysql_tests.erl, line 278)
**error:{badmatch,{error,{1364,<<"HY000">>,
                  <<"Field 'tx' doesn't have a default value">>}}}
  output:<<"">>

mysql_tests: query_test_ (Binary protocol)...*failed*
in function mysql_tests:binary_protocol/1 (test/mysql_tests.erl, line 298)
**error:{badmatch,{error,{1050,<<"42S01">>,<<"Table 't' already exists">>}}}
  output:<<"">>

mysql_tests: query_test_ (INT)...*failed*
in function mysql_tests:write_read_text_binary/5 (test/mysql_tests.erl, line 527)
in call from mysql_tests:int/1 (test/mysql_tests.erl, line 412)
**error:{badmatch,{error,{1264,<<"22003">>,
                  <<"Out of range value for column 'i' at row 1">>}}}
  output:<<"">>

mysql_tests: query_test_ (DATE)...*failed*
in function mysql_tests:write_read_text_binary/5 (test/mysql_tests.erl, line 527)
in call from lists:foreach/2 (lists.erl, line 1337)
in call from mysql_tests:date/1 (test/mysql_tests.erl, line 439)
**error:{badmatch,{error,{1292,<<"22007">>,
                  <<"Incorrect date value: '0000-00-00' for column 'd"...>>}}}
  output:<<"">>

mysql_tests: query_test_ (DATETIME)...*failed*
in function mysql_tests:write_read_text_binary/5 (test/mysql_tests.erl, line 527)
in call from lists:foreach/2 (lists.erl, line 1337)
in call from mysql_tests:datetime/1 (test/mysql_tests.erl, line 473)
**error:{badmatch,{error,{1292,<<"22007">>,
                  <<"Incorrect datetime value: '0000-00-00 00:00:00' "...>>}}}
  output:<<"">>

mysql_tests: query_test_ (Microseconds)...*failed*
[...]

PR #45, opened 2016-07-23
Explicitly set the SQL mode to a known value

johlo:
To avoid unexepected differences in between MySQL versions
and distributions.

This fixes the failing unit tests in issue #43


zuiderkwast:
Thanks!

PR #47, opened 2016-08-26
Add support for the MySQL JSON data type

johlo:
This commit adds support for the JSON data type introduced
in MySQL 5.7.8.

The unit test for JSON is very fragile as it compares strings
directly. It should use a proper JSON parser and compare the
parsed terms.

### DIFF
--- a/include/protocol.hrl
+++ b/include/protocol.hrl
@@ -113,6 +113,7 @@
 -define(TYPE_YEAR, 16#0d).
 -define(TYPE_VARCHAR, 16#0f).
 -define(TYPE_BIT, 16#10).
+-define(TYPE_JSON, 16#f5).
 -define(TYPE_NEWDECIMAL, 16#f6).
 -define(TYPE_ENUM, 16#f7).
 -define(TYPE_SET, 16#f8).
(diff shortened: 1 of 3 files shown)


zuiderkwast:
Nice!

So we represent the JSON type as a binary(). That seems to make sense since there is not a consensus on how to represent JSON in Erlang and there are multiple parsers and we might not want a dependency on a JSON parser.

The test that compares the JSON as strings is maybe fragile but the imortant thing is that it doesn't pass when it shouldn't. If we wan't it less fragile, we could strip the whitespace or test it using regular expression that ignores whitespace. We can also a JSON object with only one key so there is no risk that the order would change (since the order of keys in JSON objects are undefined).

I like it so I'll merge. Thanks!

johlo:
Yes, representing it as binary() seemed like the only way to go. I didn't want to introduce a dependency on some arbitrary JSON parser or force some erlang representation on users.

Good point about test with just one key, didn't think of that :) But adding a JSON parser as a test dependency and using it to handle JSON in the tests should be fine I think, I can add that later.

I forgot to update the documentation, we should add a JSON row to the "Value Representation" table. I can open a PR later this week if you don't have time. 
@zuiderkwast

zuiderkwast:
True. I thought about adding it to the value representation table but I decided not to comment on it. :-)

For the test, I think it's fine to just compare the strings. If it breaks in a future MySQL version, we can deal with it then, but I don't think it's likely that it will change.

Thanks for your efforts! I don't have much time for this nowdays.

johlo:
You've done a good job with this library, it's the best mysql driver around, thanks for that! :)

I have some more PR's in mind for the future (e.g. the logging of slow/un-indexed queries and better travis  testing (with different mysql versions)), so I'll be pinging you again soonish ;)

PR #49, opened 2016-08-27
Update erlang.mk

johlo:
And hide new .d file generated by erlang.mk.
Also, make sure git doesn't print diff for erlang.mk when it has been updated.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+erlang.mk -diff
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ cover
 *.coverdata
 tests.output
 test/ct.cover.spec
+mysql.d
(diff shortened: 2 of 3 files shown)


johlo:
ping @zuiderkwast

zuiderkwast:
The new file is very large. Potentially there is a risk that a commit like this one could do something unexpected, I mean, when the diff is hidden or too large. I don't think it is very likely now but it annoys me a little.

Is there a specific reason to update Erlang.mk? Any new features or bug fixes that we need?

There is a way to build Erlang.mk with just the features that we need, by putting a config file in the repo. Maybe there should be such an erlang.mk config file included in our repo....

johlo:
No, I don't have any issues with the current erlang.mk that is in the repo. I like keeping dependencies up-to-date to avoid major incompatibilities in the future.

I did the update by running `make erlang-mk`. This updates erlang.mk to the `2.0.0-pre.2` tag, which is the latest release.

I don't see any major changes in the functionality in erlang.mk, mostly bugfixes. The size explosion comes from the updates to the metadata for the packages that is maintained in the erlang.mk file.

Feel free to close this PR without merging if you don't like it, it works as is today.

zuiderkwast:
I have added a `erlang-mk.build.config` for a custom (small) erlang.mk build which shall be used for future updates.

I have also updated erlang.mk using this custom build.

PR #56, opened 2017-01-24
Allow setting CLIENT_FOUND_ROWS on handshake.

kzemek:
The option is useful for compatibility with other SQL databases and
drivers (ODBC, PostgreSQL).

### DIFF
--- a/include/protocol.hrl
+++ b/include/protocol.hrl
@@ -26,6 +26,9 @@
 
 %% --- Capability flags ---
 
+%% Server: sends found rows instead of affected rows in EOF_Packet
+-define(CLIENT_FOUND_ROWS, 16#00000002).
+
 %% Server: supports schema-name in Handshake Response Packet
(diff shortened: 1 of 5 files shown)


zuiderkwast:
Hi kzemek! I think this is useful. Just add some docs and tests. :-)

1. The name `set_found_rows_cap` contains implementation details. The user shouldn't need to know what a capability flag is. How about just `found_rows`?
1. A test case where the result of affected_rows/1 would be different depending on this option.
1. Documentation of the option in connect. Something like "Makes affected_rows/1 return the number of found (matched) rows rather than the number of changed rows (for UPDATE queries)".
1. Also mention the parameter in the docs for affected_rows/1.

Thanks!

kzemek:
I'm not sure about the third point; I've already added documentation for `mysql:start_link/1`, where is the other place I would document the new option in?

zuiderkwast:
Right, docs for start_link/1, that's what I meant.

Great! Thanks!

PR #58, opened 2017-02-07
Fix reading of unsigned integers in binary protocol

weisslj:
Before this patch, e.g. -1536 was returned instead of 64000 for small unsigned integers.


zuiderkwast:
Looks correct to me. Sorry for the long delay. Merged now. Thanks!

PR #59, opened 2017-02-07
Fix tests for MariaDB

weisslj:


zuiderkwast:
I have not tested this but it I trust you that this is correct. Thanks!

PR #61, opened 2017-02-27
Fix parsing of unsigned integers

imrivera:


imrivera:
I see that there is already another previous pull request fixing this problem. I'll close this one.

zuiderkwast:
Yes, #58. This one is looks pretty much the same. Thanks for your effort anyway!

PR #62, opened 2017-03-02
Allow configurable logging of errors. Default is set to error_logger

slezakattack:


zuiderkwast:
This is interesting.

On one hand, I like to do things the OTP way. I also like to keep configuration somewhat minimal. Error_logger is the standard interface to error logging in OTP. "The standard behaviours (supervisor, gen_server, and so on) send progress and error information to error_logger." (from the docs of error_logger)

I'm aware that lager has throttling and there is some more alternative to lager too. Lager can also hook in as an even handler for error_logger, and the standard event handler for logging to disk can be disabled, so error_logger can still be used as the logging interface without the writing-to-disk bottleneck. See this thread: https://stackoverflow.com/questions/33992949/erlang-error-logger-and-overload-protection

The other problem is that the process message inbox can be flooded. If the OTP way is bad, maybe there is some improvement in OTP going on. From the release notes of 19.3 http://erlang.2086793.n4.nabble.com/Erlang-OTP-19-3-has-been-released-td4719987.html I read this:

> kernel: Two OTP processes that are known to receive many
> messages are rex (used by rpc) and error_logger.
> Those processes will now store unprocessed messages
> outside the process heap, which will potentially
> decrease the cost of garbage collections.

Are you fine with this? If not, can you explain what the pluggable logging fun can do that is not possible to achieve with the other approaches?

slezakattack:
Hey Viktor, thanks for the reply. I admit that the solution I'm providing adds a bit more weight to the library. I actually wasn't aware about disabling the logfile for error_logger. Given that, you can reject this PR as I'll just disable error_logger's file logging. Thanks!

zuiderkwast:
OK, great!

PR #65, opened 2017-04-13
fix mysql connect timeout.

hunkhl:
mysql 5.7.17
when mysql client connect to mysql server success, mysql server will push handshake data immediately.
fix sometimes mysql_otp client may lost receive handshake data.


zuiderkwast:
Makes sense. Merged. Thanks!

PR #67, opened 2017-05-11
erlang 20 add floor/1 function, change the floor/1 funcion name in self create module.

getong:


gogocurtis:
+1

gogocurtis:
I'm using ERLC_OPTS=-W0 mix deps.compile ( elixir ) to work around this issue in Erlang20

zuiderkwast:
Thanks!

PR #68, opened 2017-05-21
Add support for auth method switch

fenek:
This PR adds support for `AuthSwitchRequest` packet that can be returned by the server as a result to handshake response. This is required e.g. to connect to managed MySQL instance in Microsoft Azure, which doesn't allow "fast authentication path".

Most probably this improvement doesn't cover all possible paths but is sufficient to connect to Azure's MySQL.

### DIFF
--- a/include/records.hrl
+++ b/include/records.hrl
@@ -27,6 +27,11 @@
                     auth_plugin_data :: binary(),
                     auth_plugin_name :: binary()}).
 
+-record(auth_method_switch, {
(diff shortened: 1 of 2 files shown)


fenek:
Any news on this? :) If it's about the tests, I'm not sure how to configure MySQL to send this packet after handshake. :(

zuiderkwast:
Seems to be OK. Thanks!

PR #69, opened 2017-06-02
SSL connection support

fenek:
This PR adds SSL connection support. SSL can be enabled by adding `{ssl, [ssl:ssl_option()]}` connection option.

Thanks @michalslaski for coop. :)

### DIFF
--- a/src/mysql_sock_ssl.erl
+++ b/src/mysql_sock_ssl.erl
@@ -1,0 +1,64 @@
+%% MySQL/OTP – MySQL client library for Erlang/OTP
+%% Copyright (C) 2017 Piotr Nosek, Michal Slaski
+%%
+%% This file is part of MySQL/OTP.
+%%
+%% MySQL/OTP is free software: you can redistribute it and/or modify it under
+%% the terms of the GNU Lesser General Public License as published by the Free
+%% Software Foundation, either version 3 of the License, or (at your option)
+%% any later version.
+%%
+%% This program is distributed in the hope that it will be useful, but WITHOUT
+%% ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+%% FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+%% more details.
+%%
+%% You should have received a copy of the GNU Lesser General Public License
+%% along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+%% @doc This module provides SSL socket interface, i.e. is a proxy to ssl module.
+%% @private
+-module(mysql_sock_ssl).
+
+-export([connect/3, close/1, send/2, recv/2, recv/3]).
+-export([setopts/2]).
+
+%% --------------------------------------------------
+%% API
+%% --------------------------------------------------
+
+connect(Port, ConfigSSLOpts, Timeout) ->
+    DefaultSSLOpts = [{versions, [tlsv1]}, {verify, verify_peer}],
+    MandatorySSLOpts = [{active, false}],
+    MergedSSLOpts = merge_ssl_options(DefaultSSLOpts, MandatorySSLOpts, ConfigSSLOpts),
+    ssl:connect(Port, MergedSSLOpts, Timeout).
+
+close(Socket) ->
+    ssl:close(Socket).
+
+send(Socket, Packet) ->
+    ssl:send(Socket, Packet).
+
+recv(Socket, Length) ->
+    ssl:recv(Socket, Length).
+
+recv(Socket, Length, Timeout) ->
+    ssl:recv(Socket, Length, Timeout).
+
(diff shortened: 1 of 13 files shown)


fenek:
Actually please hold on, as I've found out recently, that this has to be improved:

https://github.com/mysql-otp/mysql-otp/pull/69/files#diff-4de76a68a0a262b3cd7a6e54af573536R250

It seems that MySQL doesn't use client capabilities announced after SSL negotation, but sticks to the original ones so we can't simply pass `(false, false)` here.

I will commit a patch soon.

fenek:
OK, done. :) Ready for review. :)

fenek:
I'm getting random `closed` error from server. :( Is it known issue with Travis environment or is my PR breaking something? :(

zuiderkwast:
I will look at this PR in a few weeks. I'm travelling atm. It's a useful feature indeed!

fenek:
Hey. :) It would really help us move forwards with MongooseIM roadmap if we could get this PR merged. :) Is there anything I can help you with? Are these random Travis failures something new and introduced by this PR?

zuiderkwast:
Sorry for the delay! I have been traveling. I will look at it now. I don't know why Travis seems to fail randomly.

src/mysql_sock_ssl.erl
```diff
@@ -0,0 +1,64 @@
+%% MySQL/OTP – MySQL client library for Erlang/OTP
+%% Copyright (C) 2014 Viktor Söderqvist
```
zuiderkwast: Please use your own name and the current year for these new files. :-) Also feel free to add your own name to old files where major changes have been done. The year should probably be updated too.

src/mysql_sock_tcp.erl
```diff
@@ -0,0 +1,42 @@
+%% MySQL/OTP – MySQL client library for Erlang/OTP
+%% Copyright (C) 2014 Viktor Söderqvist
```
zuiderkwast: Use your own name and the current year.

test.ssl/Makefile
```diff
@@ -0,0 +1,27 @@
+.PHONY: all clean
```
zuiderkwast: Can these files be placed in e.g. test/ssl/ so that all test related files are under test/ ? Less cluttered directory structure IMO.

src/mysql_sock_ssl.erl
fenek: Good, old copy & paste problem. :) I will fix it.

test.ssl/Makefile
fenek: Of course, will change it.

fenek:
Have you seen these random failures in other PRs? Or is it just this one?

zuiderkwast:
I have only seen the random failures on this PR.

fenek:
Thanks. I will try to find out what might be wrong.

fenek:
It seems it was all because of connection limit. Perhaps some connections in tests are not closed in clean way? I guess it's a matter of another PR.

I've also updated docs. I forgot to do it earlier. :(

getong:
@fenek 
```
root@myhost [(none)]> set global ssl_ca = '/etc/ssl/certs/ca-bundle.crt';
ERROR 1238 (HY000): Variable 'ssl_ca' is a read only variable
root@myhost [(none)]> set global ssl_cert = '/etc/ssl/somepath/mysql.crt' ;
ERROR 1238 (HY000): Variable 'ssl_cert' is a read only variable
root@myhost [(none)]> set global ssl_key = '/etc/ssl/somepath/private/mysql.key';
ERROR 1238 (HY000): Variable 'ssl_key' is a read only variabl
```
See https://bugs.mysql.com/bug.php?id=81461
I think the last commit is wrong. And the first two commits are right but somewhere need to fix.

fenek:
Yes, the last commit is not the right way to go. Most certainly the best one so far is adding SSL settings to `conf.d` but it was still not enough. Will keep debugging it today... so sorry for lots of notifications about updated PR. :)

Unfortunately it's most probably some weirdness of Travis environment that needs a walk-around...

getong:
@fenek You are doing the right thing, keep moving on.

fenek:
@getong @zuiderkwast 

It seems I have found the cause of these failures - there was a race condition, where a message about connection process exit from basic tests could leak into keepalive test case. Now the connection close is synchronous, so the problem has disappeared.

I think the PR is ready for merge, unless you have some other comments? :)

fenek:
Will you have some spare time soon to check it? :) We would really like to use this version in our next MongooseIM release. :)

zuiderkwast:
I guess this is good now. :-) Sorry for not looking until now. I have been travelling in offline countries now for a while.

Thanks!

Issue #72, opened 2017-07-10
MariaDB 10.2.6 error in handshake


ea7ababe:
Hello there,

today I have updated MariaDB from 10.1.24 to 10.2.6 and now `mysql:start_link([{name, {local,db}},{database,D},{user,U},{password,P}])` raises an exception:

```
** exception exit: {badmatch,2}
     in function  mysql_protocol:handshake/6 (/usr/home/u/project/_build/default/lib/mysql/src/mysql_protocol.erl, line 60)
     in call from mysql:init/1 (/usr/home/u/project/_build/default/lib/mysql/src/mysql.erl, line 497)
     in call from gen_server:init_it/6 (gen_server.erl, line 328)
     in call from proc_lib:init_p_do_apply/3 (proc_lib.erl, line 247)
```

When I comment out that line 60 in mysql_protocol.erl everything seems to work fine...

ionuthristodorescu:
Same here, on macOS Sierra 10.12.6. Has anybody been able to look into this issue ? It fails out this assert at line 60 : "OkStatus = Handshake#handshake.status" because OkStatus is 16386 (0x4002 in hex) and handshake.status is 2.

ionuthristodorescu:
So, looks like for Status fields (https://dev.mysql.com/doc/dev/mysql-server/8.0.0/mysql__com_8h.html), 0x4002 is actually SERVER_STATUS_AUTOCOMMIT | SERVER_SESSION_STATE_CHANGED. The problem is, SERVER_SESSION_STATE_CHANGED should be set in the Status flags ONLY if CLIENT_SESSION_TRACK is set - but it says in the comments in mysql_protocol.erl that "We have CLIENT_PROTOCOL_41 but not CLIENT_SESSION_TRACK enabled". I'll continue debugging and update here.

ionuthristodorescu:
https://mariadb.com/kb/en/mariadb/1-connecting-connecting/  : 

if (server_capabilities & CLIENT_MYSQL)
  string<4> filler
else
  int<4> server capabilities 3rd part . MariaDB specific flags /* MariaDB 10.2 or later */

sfilargi:
Is there a workaround?

sfilargi:
Should we be looking at the client flags, or the server flags?
I.e.

```
#define SERVER_STATUS_ANSI_QUOTES       32768U
```

sfilargi:
[...]

PR #75, opened 2017-12-18
Compatible with MySQL 5.7 and MariaDB 10.2.6+

emqplus:
Fix issue https://github.com/emqtt/emq-auth-mysql/issues/62
Fix issue https://github.com/mysql-otp/mysql-otp/issues/72

### DIFF
--- a/src/mysql_protocol.erl
+++ b/src/mysql_protocol.erl
@@ -72,7 +72,8 @@ handshake_finish_or_switch_auth(Handshake, Password, SockModule, Socket,
     {ok, ConfirmPacket, SeqNum1} = recv_packet(SockModule, Socket, SeqNum0),
     case parse_handshake_confirm(ConfirmPacket) of
         #ok{status = OkStatus} ->
-            OkStatus = Handshake#handshake.status,
+            true = (OkStatus == Handshake#handshake.status)
+                    orelse (OkStatus == (16#4000 bor Handshake#handshake.status)),
             {ok, Handshake, SockModule, Socket};
         #auth_method_switch{auth_plugin_name = AuthPluginName,
                             auth_plugin_data = AuthPluginData} ->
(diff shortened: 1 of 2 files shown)


src/mysql_protocol.erl
```diff
@@ -72,7 +72,8 @@ handshake_finish_or_switch_auth(Handshake, Password, SockModule, Socket,
     {ok, ConfirmPacket, SeqNum1} = recv_packet(SockModule, Socket, SeqNum0),
     case parse_handshake_confirm(ConfirmPacket) of
         #ok{status = OkStatus} ->
-            OkStatus = Handshake#handshake.status,
+            true = (OkStatus == Handshake#handshake.status)
+                    orelse (OkStatus == (16#4000 bor Handshake#handshake.status)),
```
zuiderkwast: 1. Please add a comment before these two lines. Suggestion:

```erlang
            %% check status, ignoring bit 16#4000, SERVER_SESSION_STATE_CHANGED.
```

2. Why not
```erlang
            true = OkStatus band bnot 16#4000 == Handshake#handshake.status,
```

I think it is more ideomatic to my understanding of it: _match, ignoring bit 16#4000" in OkStatus_

emqplus: @zuiderkwast Cannot match a record field in Erlang. I updated the PR.

```
ERLC   mysql_protocol.erl
src/mysql_protocol.erl:77: illegal pattern
make[1]: *** [ebin/mysql.app] Error 1
make: *** [app] Error 2
```

zuiderkwast:
Thanks!

emqplus:
@zuiderkwast Happy new year. Could we tag 1.3.1 for the PR? We are planning to release EMQ 2.3.3 which depends on the latest mysql-otp.

zuiderkwast:
Yes.... will do

PR #77, opened 2018-01-21
Add capability of sending packets with size greater than 16#ffffff

skaee:
A little add-on marked as Todo in the code:
- Add capability of sending packets with size greater than 16#ffffff

### DIFF
--- a/src/mysql_protocol.erl
+++ b/src/mysql_protocol.erl
@@ -1018,11 +1018,18 @@ parse_packet_header(<<PacketLength:24/little-integer, SeqNum:8/integer>>) ->
 add_packet_headers(PacketBody, SeqNum) ->
     Bin = iolist_to_binary(PacketBody),
     Size = size(Bin),
-    SeqNum1 = (SeqNum + 1) band 16#ff,
-    %% Todo: implement the case when Size >= 16#ffffff.
-    if Size < 16#ffffff ->
-        {[<<Size:24/little, SeqNum:8>>, Bin], SeqNum1}
-    end.
+    add_packet_headers(Bin, Size, <<>>, SeqNum).
+
+%% @doc Manages all sizes of datas to send even with sizes greater than 0xffffff (16#ffffff).
+-spec add_packet_headers(PacketBody :: iodata(), Size :: integer(), Acc :: iodata(), SeqNum :: integer()) ->
+    {PacketWithHeaders :: iodata(), NextSeqNum :: integer()}.
+add_packet_headers(Bin, Size, Acc, SeqNum) when Size > 16#ffffff ->
+    NextSeqNum = (SeqNum + 1) band 16#ff,
+    << Payload:16#ffffff/binary, Rest/binary >> = Bin,
+    add_packet_headers(Rest, Size - 16#ffffff, << Acc/binary, 16#ff, 16#ff, 16#ff, SeqNum:8, Payload/binary >>, NextSeqNum);
+add_packet_headers(Bin, Size, Acc, SeqNum) ->
+    NextSeqNum = (SeqNum + 1) band 16#ff,
+    { [Acc, <<Size:24/little, SeqNum:8>>, Bin], NextSeqNum}.
 
 -spec parse_ok_packet(binary()) -> #ok{}.
 parse_ok_packet(<<?OK:8, Rest/binary>>) ->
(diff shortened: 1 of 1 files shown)


src/mysql_protocol.erl
```diff
@@ -1259,6 +1266,23 @@ add_packet_headers_test() ->
     {Data, 43} = add_packet_headers(<<"foo">>, 42),
     ?assertEqual(<<3, 0, 0, 42, "foo">>, list_to_binary(Data)).
 
+add_packet_headers_equal_to_0xffffff_test() ->
+	BigBin = erlang:list_to_binary([16#01 || I <- lists:seq(1, 16#ffffff)]),
+    {Data, 43} = add_packet_headers(BigBin, 42),
+    ?assertEqual(<<16#ff, 16#ff, 16#ff, 42, BigBin/binary>>, list_to_binary(Data)).
```
zuiderkwast: This is not the correct behaviour. A packet size of 16#ffffff means that one more packet will come. There is no other indicator of if there are more packets to come or not. Thus, a packet of size 0 has to follow. See https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_packets.html .

In the implementation, it has to be `>=` in the guard, not `>`.

src/mysql_protocol.erl
```diff
@@ -1018,11 +1018,18 @@ parse_packet_header(<<PacketLength:24/little-integer, SeqNum:8/integer>>) ->
...
-    %% Todo: implement the case when Size >= 16#ffffff.
-    if Size < 16#ffffff ->
-        {[<<Size:24/little, SeqNum:8>>, Bin], SeqNum1}
-    end.
+    add_packet_headers(Bin, Size, <<>>, SeqNum).
+
+%% @doc Manages all sizes of datas to send even with sizes greater than 0xffffff (16#ffffff).
+-spec add_packet_headers(PacketBody :: iodata(), Size :: integer(), Acc :: iodata(), SeqNum :: integer()) ->
+    {PacketWithHeaders :: iodata(), NextSeqNum :: integer()}.
+add_packet_headers(Bin, Size, Acc, SeqNum) when Size > 16#ffffff ->
```
zuiderkwast: Why not use `byte_size(Bin)` in the guard? Then the Size parameter can be removed.

src/mysql_protocol.erl
```diff
@@ -1018,11 +1018,18 @@ parse_packet_header(<<PacketLength:24/little-integer, SeqNum:8/integer>>) ->
...
     Bin = iolist_to_binary(PacketBody),
     Size = size(Bin),
-    SeqNum1 = (SeqNum + 1) band 16#ff,
-    %% Todo: implement the case when Size >= 16#ffffff.
-    if Size < 16#ffffff ->
-        {[<<Size:24/little, SeqNum:8>>, Bin], SeqNum1}
-    end.
+    add_packet_headers(Bin, Size, <<>>, SeqNum).
+
+%% @doc Manages all sizes of datas to send even with sizes greater than 0xffffff (16#ffffff).
```
zuiderkwast: This comment looks like a commit message, something for the change log. For describing the function, something in line with "helper for add_packet_headers/2" would be more useful, I think.

src/mysql_protocol.erl
```diff
@@ -1018,11 +1018,18 @@ parse_packet_header(<<PacketLength:24/little-integer, SeqNum:8/integer>>) ->
...
+-spec add_packet_headers(PacketBody :: iodata(), Size :: integer(), Acc :: iodata(), SeqNum :: integer()) ->
+    {PacketWithHeaders :: iodata(), NextSeqNum :: integer()}.
+add_packet_headers(Bin, Size, Acc, SeqNum) when Size > 16#ffffff ->
+    NextSeqNum = (SeqNum + 1) band 16#ff,
+    << Payload:16#ffffff/binary, Rest/binary >> = Bin,
+    add_packet_headers(Rest, Size - 16#ffffff, << Acc/binary, 16#ff, 16#ff, 16#ff, SeqNum:8, Payload/binary >>, NextSeqNum);
```
zuiderkwast: The size bytes could be written as `16#ffffff:24/little`, i.e. in the same way as the packet size is coded in the other function clause.

This line is too long also. Please split into 80.

src/mysql_protocol.erl
zuiderkwast: I would prefer to have the `Acc` as the last parameter, just by convention.

src/mysql_protocol.erl
zuiderkwast: Nice test cases by the way. I like them.

krant: We are also using custom patched version of mysql-otp purely because of limited packet size in upstream.

krant: Nice lib by the way. I like it :)

zuiderkwast: Good to know that more people/projects need this. I would like to merge soon. (Sorry for being picky!) :-)

qingchuwudi:
Emysql's commit [fix: when a mysql **Query** is greater than 16m, it will be crashed.](https://github.com/qingchuwudi/Emysql/commit/e831e84dd8b162bc13cea0efca666c5e1a944b9e) maybe helpful.

zuiderkwast:
Merged. Thanks! (I did some changes afterwards.)

PR #84, opened 2018-06-19
(wip)test erlang 21

getong:


src/mysql.erl
```diff
@@ -389,49 +395,46 @@ transaction(Conn, Fun, Args, Retries) when is_list(Args),
...
-                    %% Re-throw with the same trace. We'll use that in the
-                    %% final {aborted, {Reason, Trace}} in the outer level.
-                    erlang:raise(throw, {implicit_rollback, N - 1, Reason},
-                                 erlang:get_stacktrace())
-            end;
-        error:{implicit_commit, _Query} = E ->
+        case N > 0 of
+            true ->
+                case N of
+                    1 ->
```
zuiderkwast: Why change this? Maybe a guard with N > 0 would be good somewhere but the nested case doesn't look very nice.

src/mysql.erl
```diff
@@ -389,49 +395,46 @@ transaction(Conn, Fun, Args, Retries) when is_list(Args),
...
-            %% These forms for throw, error and exit mirror Mnesia's behaviour.
-            Aborted = case Class of
-                throw -> {throw, Reason};
-                error -> {Reason, erlang:get_stacktrace()};
-                exit  -> Reason
-            end,
-            {aborted, Aborted}
+            erlang:raise(error, E, Stacktrace);
+        ?WITH_STACKTRACE(Class, Reason, Stacktrace)
+        handle_other_error(Conn, Class, Reason, Stacktrace)
```
zuiderkwast: I would prefer not moving this code to a function in this PR. It doesn't seem to be necessary for OTP 21 in particular. Just do the necessary changes, i.e. the ?WITH_STACKTRACE macro. Style changes like this can be done and discussed in a separate PR.

getong:
I don't know how to continue.
The `WITH_STACKTRACE` is referenced from [rebar.hrl](https://github.com/erlang/rebar3/blob/master/src/rebar.hrl),
but it uses rebar3.
Another library is [stacktrace_compat](https://github.com/g-andrade/stacktrace_compat/). It also use rebar3.

zuiderkwast:
A simple solution would be to add this to all files where get_stacktrace is used:

```Erlang
-ifdef(OTP_RELEASE).
-compile({nowarn_deprecated_function, [{erlang, get_stacktrace, 0}]}).
-endif.
```

getong:
I don't agree with you.

zuiderkwast:
How about this then?

-ifdef(OTP_RELEASE).
-define(WITH_STACKTRACE(T, R, S), T:R -> S = erlang:get_stacktrace(),).
-else.
-define(WITH_STACKTRACE(T, R, S), T:R:S ->).
-endif.

getong:
My code is the same too, but I don't know how to use it with erlang.mk.

zuiderkwast:
I don't think it should be done added in the Makefile. When used as a dependency (which is the expected usage) the library can be built with rebar 2, rebar 3 or with make or with any other tool that understands normal OTP application file structure. There is nothing special required to build it and no dependencies except OTP itself. That's good I think. (The stuff in the Makefile is mostly used for all the other things such as building the docs and such things, not needed when used as a dependency.)

So, my suggestion is to just add it in the Erlang code and check macros already defined by OTP, e.g. OTP_RELEASE which was added in OTP 21. Possibly in an .hrl file, but why not directly in the module where it is used? It is only in one mysql module plus one module used for testing.

zuiderkwast:
There are some test cases where the connection should fail. We are testing the the error handling works as expected. The username "dummy" is used in these test cases. I think this has something to do with the changes in `error_logger` and the new `logger`.

zuiderkwast:
After a closer look, I see now that there seems to be something broken in OTP 20.3 too. Something TLS/SSL related it seems. So we have multiple issues. I have create a separate issue #85 for this error. Thanks for finding the errors!

getong:
wait for travis-ci/travis-ci#10154

src/mysql.erl
```diff
@@ -439,9 +442,6 @@ encode(Conn, Term) ->
 
 %% --- Gen_server callbacks ---
 
--include("records.hrl").
```
zuiderkwast: These include files define records and macros used for the gen_server callbacks. There are included here together with the definition of the gen_server state record. They are not used for the public API functions above. (The gen_server behaviour part should possibly be moved to it's own module in the future.)

include/records.hrl
```diff
@@ -60,3 +60,13 @@
                    orig_query :: iodata(),
                    param_count :: integer(),
                    warning_count :: integer()}).
+
+-ifdef(OTP_RELEASE). %% this implies 21 or higher
+-define(EXCEPTION(Class, Reason, Stacktrace), Class:Reason:Stacktrace).
```
zuiderkwast: Nice macros, but they don't belong in the records.hrl file. It is misleading and they are used in some files where the protocol records are not needed. Please move to a new file `exception.hrl` file or similar.

.travis.yml
```diff
@@ -14,10 +14,9 @@ before_script:
...
-  - 20.0
-  - 19.0
+  - 21.1
+  - 20.3.8.2
+  - 19.3
   - 18.3
   - 17.3
-  - 17.0
   - R16B03-1
-  - R16B
```
zuiderkwast: Why change the oldest version we support? We state in the README that we support versions from R16B. Also, I'd like to know the reason for changing 17.0 -> 17.3, 19.0 -> 19.3 and 20.0 -> 20.3.8.2.

Now, as we don't support all versions from R16B anymore, we should state in the README which versions are supported and which are not, e.g. 21.0 and 20.0-20.2(?)

src/mysql.erl
```diff
@@ -25,6 +25,9 @@
 %% gen_server is locally registered.
 -module(mysql).
 
+-include("records.hrl").
+-include("server_status.hrl").
+
 -export([start_link/1, query/2, query/3, query/4, execute/3, execute/4,
```
zuiderkwast: I'd like the public interface, i.e. the exports, to be at the top of the file. Includes are for the implementation and need not be the first thing visible in the file. See also the comment where these are moved from.

getong:
As you suggested, the code is rewritten.

PR #89, opened 2018-09-13
Fix incorrect assumption regarding lock wait timeout

emfa:
According to the documentation this has not been the default behaviour for quite some time. And the default is to not roll back on these errors. I think the problem has been masked by the extra ROLLBACK-statement.

We had problems when we got a lot of lock wait timeouts and had high load on the servers. There is a possibility of bad interleaving of processes after we incorrectly assume that the transaction bit is 0/1 and a ping-message arrives before the extra ROLLBACK is sent by the client.

Unfortunately I've not been able(or had time) to reproduce it since it requires lock waits in the DB and a high load (or the use of concuerror or the like).

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -49,7 +49,6 @@
 -define(cmd_timeout, 3000). %% Timeout used for various commands to the server
 
 %% Errors that cause "implicit rollback"
--define(ERROR_LOCK_WAIT_TIMEOUT, 1205).
 -define(ERROR_DEADLOCK, 1213).
 
 %% A connection is a ServerRef as in gen_server:call/2,3.
@@ -383,37 +382,23 @@ transaction(Conn, Fun, Args, Retries) when is_list(Args),
     %% The guard makes sure that we can apply Fun to Args. Any error we catch
     %% in the try-catch are actual errors that occurred in Fun.
     ok = gen_server:call(Conn, start_transaction),
+    execute_transaction(Conn, Fun, Args, Retries).
+
+execute_transaction(Conn, Fun, Args, Retries) ->
     try apply(Fun, Args) of
         ResultOfFun ->
-            %% We must be able to rollback. Otherwise let's crash.
             ok = gen_server:call(Conn, commit),
             {atomic, ResultOfFun}
     catch
-        throw:{implicit_rollback, N, Reason} when N >= 1 ->
-            %% Jump out of N nested transactions to restart the outer-most one.
-            %% The server has already rollbacked so we shouldn't do that here.
-            case N of
-                1 ->
-                    case Reason of
-                        {?ERROR_DEADLOCK, _, _} when Retries == infinity ->
-                            transaction(Conn, Fun, Args, infinity);
-                        {?ERROR_DEADLOCK, _, _} when Retries > 0 ->
-                            transaction(Conn, Fun, Args, Retries - 1);
-                        _OtherImplicitRollbackError ->
-                            %% This includes the case ?ERROR_LOCK_WAIT_TIMEOUT
-                            %% which we don't restart automatically.
-                            %% We issue a rollback here since MySQL doesn't
-                            %% seem to have fully rollbacked and an extra
-                            %% rollback doesn't hurt.
-                            ok = query(Conn, <<"ROLLBACK">>),
-                            {aborted, {Reason, erlang:get_stacktrace()}}
-                    end;
-                _ ->
-                    %% Re-throw with the same trace. We'll use that in the
-                    %% final {aborted, {Reason, Trace}} in the outer level.
-                    erlang:raise(throw, {implicit_rollback, N - 1, Reason},
-                                 erlang:get_stacktrace())
-            end;
+        %% We are at the top level, try to restart the transaction if there are
+        %% retries left
+        throw:{implicit_rollback, 1, _} when Retries =:= infinity ->
+            execute_transaction(Conn, Fun, Args, infinity);
+        throw:{implicit_rollback, 1, _} when Retries > 0 ->
+            execute_transaction(Conn, Fun, Args, Retries - 1);
+        throw:{implicit_rollback, N, Reason} ->
+            erlang:raise(throw, {implicit_rollback, N - 1, Reason},
+                         erlang:get_stacktrace());
         error:{implicit_commit, _Query} = E ->
             %% The called did something like ALTER TABLE which resulted in an
             %% implicit commit. The server has already committed. We need to
@@ -883,12 +868,11 @@ handle_query_call_reply([Rec|Recs], Query, State, ResultSetsAcc) ->
             ResultSetsAcc1 = [{Names, Rows} | ResultSetsAcc],
             handle_query_call_reply(Recs, Query, State, ResultSetsAcc1);
         #error{code = Code} when State#state.transaction_level > 0,
-                                 (Code == ?ERROR_DEADLOCK orelse
-                                  Code == ?ERROR_LOCK_WAIT_TIMEOUT) ->
+                                 Code == ?ERROR_DEADLOCK ->
             %% These errors result in an implicit rollback.
             Reply = {implicit_rollback, State#state.transaction_level,
                      error_to_reason(Rec)},
-            State2 = clear_transaction_status(State),
+            State2 = State#state{transaction_level = 1},
             {reply, Reply, State2};
         #error{} ->
             {reply, {error, error_to_reason(Rec)}, State}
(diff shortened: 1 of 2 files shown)


emfa:
We also hade the thought that we should disable pings while in transactions but could not decide on whether that is a good or a bad idea.

emfa:
I'll fix the tests sometime tomorrow.

zuiderkwast:
So the behaviour for the deadlocks and lock wait timeouts has changed? In which MySQL version was it changed? We claim to support MySQL 4.1 and upward. Don't we need to handle different behaviours depending on the MySQL version or some capability flag?

zuiderkwast:
The mariaex is using a test matrix with different mysql and mariadb versions combined with different OTP versions. See https://github.com/xerions/mariaex/blob/master/.travis.yml

emfa:
Yes we might need that if 4 and up should be supported... It changed in 4 if my research is correct.

emfa:
I don't know if this was a bug that was masked by the extra ROLLBACK or if that has also changed but the deadlock-error does NOT rollback the START TRANSACTION/BEGIN.

emfa:
Currently I've got no time to write a capability check though, I'll see if I got time coming week. I needed this fix since we had problems with it in production.

emfa:
If someone wants to the system variable to check for is:

`innodb_rollback_on_timeout` [https://dev.mysql.com/doc/refman/5.6/en/innodb-parameters.html#sysvar_innodb_rollback_on_timeout](url)

zuiderkwast:
Good. Is there any way to trigger a lock wait timeout? so that we can test it properly

test/transaction_tests.erl
```diff
@@ -288,8 +288,11 @@ lock_wait_timeout1({Conn1, Conn2}) ->
     {aborted, Reason} = mysql:transaction(Conn2, fun () ->
         ok = mysql:query(Conn2, "UPDATE foo SET v = 42 WHERE k = 1")
```
zuiderkwast: So, if we don't try to match this against `ok`, the transaction would not fail? I.e. it would return `{atomic, _}`?

Isn't it better to replace the `ok =` here with this ↓ then?

    ?assertMatch({error, {1205, _, <<"Lock wai", _/binary>>}},
                 mysql:query(...))`?

src/mysql.erl
```diff
@@ -383,37 +382,23 @@ transaction(Conn, Fun, Args, Retries) when is_list(Args),
...
-                    %% final {aborted, {Reason, Trace}} in the outer level.
-                    erlang:raise(throw, {implicit_rollback, N - 1, Reason},
-                                 erlang:get_stacktrace())
-            end;
+        %% We are at the top level, try to restart the transaction if there are
+        %% retries left
+        throw:{implicit_rollback, 1, _} when Retries =:= infinity ->
+            execute_transaction(Conn, Fun, Args, infinity);
+        throw:{implicit_rollback, 1, _} when Retries > 0 ->
+            execute_transaction(Conn, Fun, Args, Retries - 1);
```
zuiderkwast: When a transaction is restarted due to deadlock here, BEGIN isn't executed after the change. That means we don't get a transaction, doesn't it? Or was this an error before?

test/transaction_tests.erl
emfa: Yeah, I'll update to better reflect the new behavior.

src/mysql.erl
emfa: That was incorrect. I don't think that is anything that has been changed in any version. https://dev.mysql.com/doc/refman/5.6/en/innodb-error-handling.html

When a transaction rollback occurs due to a deadlock or lock wait timeout, it cancels the effect of the statements within the transaction. But if the start-transaction statement was START TRANSACTION or BEGIN statement, rollback does not cancel that statement.

zuiderkwast: Perfect! Thanks for this link and the quote! Can you add this quote in the a code comment?

zuiderkwast:
So, I am convinced that this is correct, at least when ` --innodb_rollback_on_timeout` is off. With this change, the application and the operator in control over the mysql configuration gets the control.

I don't really understand the problem with doing ROLLBACK and restarting the transaction though. Is it because of the high load on the servers that you think it is better to crash than to overload the servers?

The reason why I wanted to implement this logic for lock wait timeout in the first place was probably that the error message suggests it: "Lock wait timeout. Try restarting the transaction." I thought it was a pretty cool feature. I still think it makes some sense to have this functionality, but maybe there should be a limited number of retries (such as 1) and then abort? Or is it better to be dumb and let the application handle it? What do you think?

What if the `--innodb_rollback_on_timeout` is ON and a transaction just returns `{atomic, _Result}` in some case. That's not very good.  But then they should always write `ok = mysql:query(...)` to make sure the transaction is rollbacked by the exception, right?

"InnoDB rolls back only the last statement on a transaction timeout by default. If --innodb_rollback_on_timeout is specified, a transaction timeout causes InnoDB to abort and roll back the entire transaction (the same behavior as in MySQL 4.1)."

I'm ready to merge this. Just waiting to see if you have any more comments or replies to my questions.

emfa:
In a perfect world there could be some more negotiation with the server for scenarios where the database has other options then the default (it would be pretty cool) but that feels like it's another issue(you can break the driver already with wierd settings I think).

Most comments in the code seemed to suggest that it was implemented that way because of that wrong assumptions. Which makes sense because the documentation for MySQL/InnoDB is all over the place(I had to search for 10min to re-find the above link about lock wait time out...).

It could be a cool feature but then that should probably be opt-in, and accompanied with a back-off strategy of some sort. So as for now, it really needs to be up to the application. But... it does not say to restart it right away though :). Remember why this error is received in the first place; we have contention on a lock. Hammering that lock more can(and does) cause more problems, so we need to back-off.

The back-off can be put in a lot of places in peoples systems depending on their needs, not necessarily right at the database if that makes sense? Like you could back of all the way out at the client, or in the back-end, or if it was a feature here at the driver level. But I think that for reasons above that it should be an option.

Why do you want to exit the transaction fully? I felt that is was there because there had been problems (see the old comment in the code about the transaction not exiting fully). And since it is no longer needed (hopefully) I felt it best to remove it since it better follows MySQL and one RTT less per retry does not hurt either.

Awesome! Let's hope this fixes our problems. It should fix some wierd intermittent fun concurrent erlang bugs atleast. I'm still not convinced that this is what caused ours since I could not reproduce it. But we'll see soon I guess.

zuiderkwast:
Nice. I agree that backing off is the right thing to do. Thanks for the reasoning. Backing off in the application, e.g. all the way to the client using the let-it-crash principle or fallbacking to another database replica, etc.

>Most comments in the code seemed to suggest that it was implemented that way because of that wrong assumptions. Which makes sense because the documentation for MySQL/InnoDB is all over the place(I had to search for 10min to re-find the above link about lock wait time out...).

Right, so it was good to have those comments, explaining the intentions. :-) That's why I want the new quote that you found, to support the new behaviour. I can add it later if you don't.

> Why do you want to exit the transaction fully?

Now I don't, but it was needed in order to restart the transaction in the lock wait timeout case. To restart it, it first had to be rollbacked fully. That was not the case with deadlock though, but apparently I didn't know that until you showed me the quote.

emfa:
Well, a change without reason is stupid(at the very least in critical software) so no problems. :)
The comments definitely helped. Yeah, sure, add them, the more the merrier!

Yeah, it's kinda wierd behavior if you ask me but we have to live with it.

It's been running for a while now and we have not encountered the crashes we had so far, peppar peppar!

Thanks for the quick merge.

src/mysql.erl
```diff
@@ -883,12 +868,11 @@ handle_query_call_reply([Rec|Recs], Query, State, ResultSetsAcc) ->
...
             handle_query_call_reply(Recs, Query, State, ResultSetsAcc1);
         #error{code = Code} when State#state.transaction_level > 0,
-                                 (Code == ?ERROR_DEADLOCK orelse
-                                  Code == ?ERROR_LOCK_WAIT_TIMEOUT) ->
+                                 Code == ?ERROR_DEADLOCK ->
             %% These errors result in an implicit rollback.
             Reply = {implicit_rollback, State#state.transaction_level,
                      error_to_reason(Rec)},
-            State2 = clear_transaction_status(State),
+            State2 = State#state{transaction_level = 1},
```
zuiderkwast: Hey, this looks odd. The old code set transaction_level = 0 here. What's the reason for this change?

emfa: The BEGIN is not rolled back, thus we are still in level 1. Old code assumed we exited the transaction fully and hence set the level to 0.

zuiderkwast: Ah, right!

I am writing a test case for a transaction with deadlock and no retries (or all retries failing). This is not covered by any tests yet.

Issue #90, opened 2018-10-01
Another mysql:transaction-bug


emfa:
So the last fix did do a lot of good but we still have experienced some weird behaviours (although I think the root is not in the driver). 

However, I've found one more thing that have happened to us in production that might need fixing.

Since the server is oblivious to the client (we never monitor/link to the client. I'm starting to wonder if this should be fixed. In that case the server would know which processes is using it in the current transaction and could thus clean up if something happens to it).

Anyhow, the issue at hand...

The call to gen_server:call(C, start_transaction) is made without a timeout and thus if that call takes more then 5s the client dies with a timeout BUT the start_transaction might still go through. If this happens and if you are using something like poolboy this might(and have) lead to connections being returned to poolboy in an inconsistent state(still in a transaction).

I think the proper way to fix this is to just set timeout to infinity. If timeout is needed that can be imposed from somewhere else. A more elegant way would probably be something along the lines described above with proper monitoring of the clients.

zuiderkwast:
There is a macro for rollback and commit is called `?cmd_timeout` (I think) and is 3s but feel free to increase it. What do you think is appropriate? 30s?

You think for start trancaction we should use infinity? Ok, fine!

I'll accept such a PR.

It's a pity that poolboy is not linking to or monitoring the processes, as would a supervisor, but that's another project. A better pool handling would add the clients to a supervisor so that they are properly supervised, or what do you have in mind?

PR #91, opened 2018-10-05
Set gen_server-timeouts for transaction statements to infinity.

emfa:
Sorry for a pretty unclear issue, should've just done an MR straight away.

This is a freak bug that only happens in freak situations. But it wont hurt anything that I can't see and it remove one of the problems we had.

The problem was not that poolboy does not link/monitor, the exact opposite. We have client processes that checks out a db-connection and they died on the start_transaction because we had heavy load (because of another bug, see the other MR). Thus the client died with exit:timeout and poolboy reclaimed the connection, but the driver don't know that the client has died and will still faithfully execute the start_transaction. We now have a situation were there is a connection in the pool which is still in a transaction and will NEVER leave it which might cause catastrophic failure as you might imagine.

This small fix stops that from happening but it will still timeout if cmd_timeout is not met(as it should).

Hope that clears it up.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -383,13 +383,13 @@ transaction(Conn, Fun, Args, Retries) when is_list(Args),
                                            is_function(Fun, length(Args)) ->
     %% The guard makes sure that we can apply Fun to Args. Any error we catch
     %% in the try-catch are actual errors that occurred in Fun.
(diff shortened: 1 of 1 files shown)


zuiderkwast:
Thanks!

zuiderkwast:
Related to #90

PR #92, opened 2018-10-05
Dont use throw for implicit_rollback, use error as with implicit_commit.

emfa:
I would understand if this is not accepted but we need it and I could not find a good argument why throw was used.

Anyways, our problems was a try-catch that caught the implicit_rollback.

We have application code that uses throw to signal a broad range of common errors and the like. At one place we had been sloppy and caught too much, the implicit_rollback.

But as I said above, I could not find a good argument for why throw is used since this just limits the functionality of the driver when it does not have to. With this MR it is perfectly fine to catch throw:Value in the transaction function.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -394,12 +394,12 @@ execute_transaction(Conn, Fun, Args, Retries) ->
     catch
         %% We are at the top level, try to restart the transaction if there are
         %% retries left
-        ?EXCEPTION(throw, {implicit_rollback, 1, _}, _Stacktrace) when Retries =:= infinity ->
+        ?EXCEPTION(error, {implicit_rollback, 1, _}, _Stacktrace) when Retries =:= infinity ->
             execute_transaction(Conn, Fun, Args, infinity);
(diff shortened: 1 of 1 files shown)


zuiderkwast:
It may seem strange that we use throw for one thing and error for the other. I'll just try to explain why.

Throw is just described as a non-local return in the Erlang manual. It doesn't mean there is an error. As you know, we use it to retry the transaction. Unfortunately, deadlocks can happen. It doesn't mean there is a programming error.

An implicit commit (which you get when doing ALTER TABLE inside a transaction, etc.) on the other hand is a programming error that can be solved with the help of the trace.

Catching all `throw` exceptions in general, and in inside a transaction fun in particular, doesn't seem like very good style. If you need to do it, why not do it outside the transaction?

You should be careful with the code inside a transaction fun. For example, it should not have any side-effects, as it may be run multiple times in the event of a deadlocks.

Not sure if my reasoning holds. :-) What do you think?

emfa:
Good points. First of all, yes, it's very bad practice, or you should atleast be alot more careful that it was done in this particular case :) I'm convinced by your arguments and I've thought about how to approach it. Catching them outside is definitely the correct thing to do, no need to change the throw.

But that got me thinking. Check the new MR.

PR #93, opened 2018-10-07
Do not leak a throw from the API.

emfa:
If throw should be able to be used as you say, then isn't it a bit wierd that we leak the throw from the driver if all retries fail?

I propose that we add a base case that turns the throw into an error if all retries fail.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -398,6 +398,11 @@ execute_transaction(Conn, Fun, Args, Retries) ->
             execute_transaction(Conn, Fun, Args, infinity);
         ?EXCEPTION(throw, {implicit_rollback, 1, _}, _Stacktrace) when Retries > 0 ->
             execute_transaction(Conn, Fun, Args, Retries - 1);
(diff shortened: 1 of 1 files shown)


emfa:
Another argument for this is that we might want the trace to debug which statement is involved in the deadlock. This was previously lost since we threw it away for all throws.

zuiderkwast:
True, I have done this already. I though I had pushed it but I hadn't!

zuiderkwast:
I pushed it now. I have fixed the tests as well and I added a test for a deadlock without retries. Sorry for this.

I return `{aborted, {{1213, <<"40001">>, <<"Deadlock", _/binary>>}, _Trace}}`. I see you added an extra level with `{transaction_aborted,  {_,_,_}}` here. I'm not sure what is best to return. What do you think? It is undocumented for now.

emfa:
Ok, cool!

Then just disregard this. About the return value, meh, don't really care ;)

PR #96, opened 2018-11-26
Monitor application process and stop the connection in case it dies during a transaction

silviucpp:
I guess we need something like this. I tested and fixes the problem but creates lot of noise when connection is killed.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -891,14 +891,15 @@ handle_query_call_reply([], _Query, State, ResultSetsAcc) ->
         [_|_]                 -> {ok, lists:reverse(ResultSetsAcc)}
     end,
     {reply, Reply, State};
-handle_query_call_reply([Rec|Recs], Query, State, ResultSetsAcc) ->
+handle_query_call_reply([Rec|Recs], Query, #state{monitors = Monitors} = State, ResultSetsAcc) ->
     case Rec of
         #ok{status = Status} when Status band ?SERVER_STATUS_IN_TRANS == 0,
                                   State#state.transaction_level > 0 ->
             %% DDL statements (e.g. CREATE TABLE, ALTER TABLE, etc.) result in
             %% an implicit commit.
             Reply = {implicit_commit, State#state.transaction_level, Query},
-            {reply, Reply, State#state{transaction_level = 0}};
+            NewMonitors = demonitor_processes(Monitors, length(Monitors)),
+            {reply, Reply, State#state{transaction_level = 0, monitors = NewMonitors}};
         #ok{} ->
             handle_query_call_reply(Recs, Query, State, ResultSetsAcc);
         #resultset{cols = ColDefs, rows = Rows} ->
@@ -911,7 +912,8 @@ handle_query_call_reply([Rec|Recs], Query, State, ResultSetsAcc) ->
                      error_to_reason(Rec)},
             %% Everything in the transaction is rolled back, except the BEGIN
             %% statement itself. Thus, we are in transaction level 1.
-            {reply, Reply, State#state{transaction_level = 1}};
+            NewMonitors = demonitor_processes(Monitors, length(Monitors) -1),
+            {reply, Reply, State#state{transaction_level = 1, monitors = NewMonitors}};
         #error{} ->
             {reply, {error, error_to_reason(Rec)}, State}
     end.
@@ -965,3 +967,10 @@ stop_server(Reason,
                          [ConnId, Reason]),
   ok = gen_tcp:close(Socket),
   {stop, Reason, State#state{socket = undefined, connection_id = undefined}}.
+
+demonitor_processes(List, 0) ->
+    List;
+demonitor_processes([{_FromPid, MRef}|T], Count) ->
+    erlang:demonitor(MRef),
+    demonitor_processes(T, Count -1).
+
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -735,8 +735,8 @@ handle_call(start_transaction, {FromPid, _},
     State1 = update_state(Res, State),
     {reply, ok, State1#state{transaction_level = L + 1, monitors = [{FromPid, MRef} | Monitors]}};
 handle_call(rollback, {FromPid, _}, State = #state{socket = Socket, sockmod = SockMod,
-                                            status = Status, transaction_level = L,
-                                            monitors = [{FromPid, MRef}|NewMonitors]})
+                                                   status = Status, transaction_level = L,
+                                                   monitors = [{FromPid, MRef}|NewMonitors]})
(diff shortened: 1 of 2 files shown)


zuiderkwast:
Something like this, yes! But since we support some pretty old Erlang versions, we shoudn't use maps. I was thinking that only one monitor in the state should be enough. At commit/rollback, we clear the monitor even if the commit is coming from the same process as the start_transaction. It should always be from the same process if the API functions are used. Using the gen_server directly (e.g. `gen_server:call(Pid, commit)`) is not supported.

silviucpp:
hmm I'm thinking if we don't have a problem with nested transactions.. 

let's suppose inside a transaction fun you can do :

fun(Pid) ->
 spawn() -> do another transaction here inside the main one but on another process end,
end,

This won't be handled in case we use only a plain monitor_ref..

zuiderkwast:
Good point about the nested transactions! Let's use an orddict then instead of the map. It is almost never more than one or two elements anyway so this is the least overhead structure I guess.

src/mysql.erl
```diff
@@ -729,10 +733,18 @@ handle_call(start_transaction, _From,
...
     SockMod:setopts(Socket, [{active, once}]),
     State1 = update_state(Res, State),
-    {reply, ok, State1#state{transaction_level = L + 1}};
-handle_call(rollback, _From, State = #state{socket = Socket, sockmod = SockMod,
-                                            status = Status, transaction_level = L})
+    {reply, ok, State1#state{transaction_level = L + 1, monitors = maps:put(FromPid, MRef, Monitors)}};
+handle_call(rollback, {FromPid, _}, State = #state{socket = Socket, sockmod = SockMod,
+                                            status = Status, transaction_level = L, monitors = Monitors})
   when Status band ?SERVER_STATUS_IN_TRANS /= 0, L >= 1 ->
+    NewMonitors = case maps:take(FromPid, Monitors) of
```
zuiderkwast: If one process does a nested transaction, we shouldn't demonitor when the inner transaction finishes. Perhaps we can use a simple list of monitors where we add one regardless of whether there already is one for the process. Then we pop them and demonitor in commit/rollback. An other solution would be to add a counter per process and demonitor only when the counter reaches zero. (I.e. `[{Pid, {Count, MonitorRef}}]` if we use an orddict.)

It's the same code for rollback and commit, so it should be moved to a helper function.

Note the recommended line length of 80 as well.

silviucpp: Yes I'll keep under loop all your observations. List works ok.

I'll do the changes and when you give me the ok I'll write the test case as well.

silviucpp: How do you prefer:

Solution 1:

- In the monitor list we add: `[{FromPid, MRef} | Monitors]}`
- in rollback and commit we do:

`handle_call(rollback, {FromPid, _}, State = #state{..., monitors = [{FromPid, MRef}|NewMonitors]})`

Solution 2:

- In the monitor list we add: `[Mref | Monitors]}`
- in rollback and commit we do:
`handle_call(rollback, _From, State = #state{..., monitors = [MRef|NewMonitors]})`

Idea is that having PID in the list it's only as a additional safe check that our logic is ok .. nothing more..

zuiderkwast: I like the Pid. It causes a crash if the wrong pid is calling commit/rollback, which is good.

Perhaps we can get rid of transaction_level. We can check if Monitors is empty instead.

silviucpp: Please check and let me know if you want to to the transaction_level change as well.

src/mysql.erl
```diff
@@ -729,10 +733,13 @@ handle_call(start_transaction, _From,
                                                ?cmd_timeout),
     SockMod:setopts(Socket, [{active, once}]),
     State1 = update_state(Res, State),
-    {reply, ok, State1#state{transaction_level = L + 1}};
-handle_call(rollback, _From, State = #state{socket = Socket, sockmod = SockMod,
-                                            status = Status, transaction_level = L})
+    {reply, ok, State1#state{transaction_level = L + 1, monitors = [{FromPid, MRef} | Monitors]}};
+handle_call(rollback, {FromPid, _}, State = #state{socket = Socket, sockmod = SockMod,
+                                            status = Status, transaction_level = L,
+                                            monitors = [{FromPid, MRef}|NewMonitors]})
```
zuiderkwast: Indentation is broken here because FromPid was added.

silviucpp: Fixed

src/mysql.erl
zuiderkwast: I see that the level is used to generate breakpoint names. We can keep it.

zuiderkwast:
I think it looks good! How about the existing tests? Some test is failing because of this monitor. Does the caller exit when the connection is in a transaction?

silviucpp:
I see all tests in trevis fails with `Access denied for user 'dummy'@'localhost` . I'll try to run tests on my machine and see.

silviucpp:
> I see all tests in trevis fails with `Access denied for user 'dummy'@'localhost` . I'll try to run tests on my machine and see.

My bad ! I ses this is an expected stuff based on the code. But I honestly don't know how to read what test failed..

silviucpp:
OK found how to interpret the errors. Need to check seems something related to implicit_commit

zuiderkwast:
The error from error_logger are trapped and compared against the expected errors, but this doesn't seem to work properly in OTP 21 because of the new logger. Look at any build for OTP 16..20 to see what failed.

zuiderkwast:
We should probably demonitor everything when implicit_commit occurs.

silviucpp:
And for implicit_rollback e demonitor all excepting the last one ? I mean the one that did the BEGIN ?

zuiderkwast:
> And for implicit_rollback e demonitor all excepting the last one ? I mean the one that did the BEGIN ?

Yes, that is correct. The transaction is retried by the one that did BEGIN. If there are no tries left, there will one that did BEGIN will call rollback which removes the last monitor.

zuiderkwast:
I think that it should always be `transaction_level == length(monitors)`.

silviucpp:
All tests are passing now. BTW tests are failing on MySQL 8 (tested on my machine but also the tests from master are failing).

silviucpp:
I also added a test. let me know what else it's needed in order to have this merged

silviucpp:
Maybe you can take a look as well why tests are failing on MySQL 8. For me
fails in master as well and I’m not sure I configured test env wrong or
something changed in MySQL.

zuiderkwast:
For MySQL 8 I guess there is a new default authentication method; one that we haven't implemented. It fails with and without this PR, right?

silviucpp:
Hello,
If fails without this PR and also is not related to that auth method :) because I added in config to use the old auth method. And all tests are passing excepting one (I don't remember now which one)

PR #98, opened 2018-12-02
Clarification for usage with MySQL 8

zuiderkwast:
* Add notes in README about unsupported caching_sha2_password.
* Explicitly use mysql_native_password in tests.
* Remove NO_AUTOC_REATE_USER from SQL_MODE as it doesn't exist anymore.
* Deadlock test still failing with MySQL 8.0.13. Implicit commit happens unexpectedly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,10 @@ before_script:
   - cat test/ssl/my-ssl.cnf | sudo tee -a /etc/mysql/conf.d/my-ssl.cnf
   - sudo service mysql start
   - sleep 5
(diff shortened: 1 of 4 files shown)


silviucpp:
Hello,

Yes this is what's very strange "Deadlock test still failing with MySQL 8.0.13. Implicit commit happens unexpectedly.".

This one it's what I observed as well when I reported that MySQL 8  tests are failing

zuiderkwast:
Yes, it's very strange. Perhaps it's a bug in MySQL 8. Perhaps they don't tests for deadlock scenarios like this one.

Issue #38, opened 2016-03-07
Support LOAD DATA LOCAL INFILE


redink:
Hello,
First, thanks for this repo, I replace mysql driver via `mysql-otp`. But, I found this driver can not support `LOAD DATA LOCAL INFILE`, just like:

```
LOAD DATA LOCAL INFILE '/path/myfile' INTO TABLE table_name;
```

is it possible, can you realize support for `LOAD DATA LOCAL INFILE`?

Thanks again.

zuiderkwast:
It would be a nice feature. I don't have time to implement it myself at the moment but feel free to implement it and submit a pull request.

PR #103, opened 2019-01-10
Support LOAD LOCAL INFILE

stuart-thackray:
for issue https://github.com/mysql-otp/mysql-otp/issues/38


zuiderkwast:
Is this all? If so, that's great, but please add a test case!

stuart-thackray:
You have to configure the Server (local_infile=1) and you have to make two calls to mysql the 1st is the instruction and then next one is the file contents; I will demonstrate with a test case.

zuiderkwast:
Sounds good. I assume we need to have `local_infile=1` in the .travis configuration to make sure the tests can run on Travis.

stuart-thackray:
Hi @zuiderkwast I was mistaken with all that was required was just the protocol was enabling the allowed protocol; I have since updated; let me know if it should be done in a diffrent way; I have included test cases also in a sperate commit. 

Let me know if the load file should be done in a diffrent way.

README.md
```diff
@@ -133,7 +133,7 @@ because CA certificates will no longer match.
 If you run `make tests COVER=1` a coverage report will be generated. Open
 `cover/index.html` to see that any lines you have added or modified are covered
 by a test.
-
+
```
zuiderkwast: Please don't add trailing whitespace

README.md
```diff
@@ -161,7 +161,7 @@ Tagging a new version:
   * Update the online documentation and coverage reports using `make gh-pages`.
     Then push the gh-pages branch using `git push origin gh-pages`.
 
-License
+License
```
zuiderkwast: Please don't add trailing whitespace

README.md
```diff
@@ -111,7 +111,7 @@ The encode and protocol test suites does not require a
 running MySQL server on localhost.
 
 For the suites `mysql_tests`, `ssl_tests` and `transaction_tests` you need to
-start MySQL on localhost and give privileges to the user `otptest` and (for
+start MySQL on localhost and give privileges to the user `otptest` additionally you need to set my.cnf with `local_infile=1`. And (for
 `ssl_tests`) to the user `otptestssl`:
```
zuiderkwast: This sentence is weird now. Better add your Additionally as a separate sentence after this one.

README.md
```diff
@@ -111,7 +111,7 @@ The encode and protocol test suites does not require a
 running MySQL server on localhost.
 
 For the suites `mysql_tests`, `ssl_tests` and `transaction_tests` you need to
-start MySQL on localhost and give privileges to the user `otptest` and (for
+start MySQL on localhost and give privileges to the user `otptest` additionally you need to set my.cnf with `local_infile=1`. And (for
```
zuiderkwast: This sentence is weird now. Better add your Additionally as a separate sentence after this one.

src/mysql_protocol.erl
```diff
@@ -423,6 +440,8 @@ fetch_response(SockModule, Socket, Timeout, Proto, Acc) ->
                     parse_ok_packet(Packet);
                 ?error_pattern ->
                     parse_error_packet(Packet);
+                <<251:8, FileName/binary>> ->
```
zuiderkwast: Please add a macro for 251 where the other packet tags are defined, with an appropriate name and comment if needed.

test/mysql_tests.erl
```diff
@@ -356,6 +358,36 @@ text_protocol(Pid) ->
 
     ok = mysql:query(Pid, <<"DROP TABLE t">>).
 
+
+load_data_file_infile(Pid) ->
+        ok = mysql:query(Pid, <<"CREATE TABLE `otptest`.`loadFile` (
+  `col1` BIGINT(20) NOT NULL,
```
zuiderkwast: Indentation looks strange in this function. Please use 4 spaces and no tabs.

src/mysql_protocol.erl
```diff
@@ -115,14 +115,30 @@ ping(SockModule, Socket) ->
 query(Query, SockModule, Socket, Timeout) ->
     Req = <<?COM_QUERY, (iolist_to_binary(Query))/binary>>,
     SeqNum0 = 0,
-    {ok, _SeqNum1} = send_packet(SockModule, Socket, Req, SeqNum0),
-    fetch_query_response(SockModule, Socket, Timeout).
+    {ok, SeqNum1} = send_packet(SockModule, Socket, Req, SeqNum0),
+    QryRsp = fetch_query_response(SockModule, Socket, Timeout),
+    case QryRsp of
+    {ok, [#load_local_file{filename = FileName}]} ->
```
zuiderkwast: Indentation after `of` please.

The naming of `QryRsp` is a bit obscure, compared to the other variable names used in this file. How about `Response` or `Resp`? No need to mention that it is from a query, since we are a function called `query`.

stuart-thackray:
Hi @zuiderkwast I was using gitkraken and was doing line by line commits; and accidently included that swp file. 

Not sure if you use linux/vim; but you can add `% vim: sw=4 ts=4 et ft=erlang` to the top of the file then once the file opened `gg=G` this will pretty print the file; or just to add it if you want to allow contirubutors to know the tab spacing etc to use. The pretty print isn't always better just sharing if you are interested.

src/mysql_protocol.erl
```diff
@@ -440,7 +441,8 @@ fetch_response(SockModule, Socket, Timeout, Proto, Acc) ->
                     parse_ok_packet(Packet);
                 ?error_pattern ->
                     parse_error_packet(Packet);
-                <<251:8, FileName/binary>> -> 
+                ?local_local_data_file ->
```
zuiderkwast: Good, but the name contains the word "local" twice. :-)

src/mysql_protocol.erl
```diff
@@ -41,6 +41,7 @@
 %% Macros for pattern matching on packets.
 -define(ok_pattern, <<?OK, _/binary>>).
 -define(error_pattern, <<?ERROR, _/binary>>).
+-define(local_local_data_file, << ?LOAD_LOCAL_FILE_REQ, _/binary >>).
```
zuiderkwast: This is macro for a pattern, so to match the others it should be named "..._pattern" like the others. But, I don't think this macro is needed. See the other comment where this is used.

src/mysql_protocol.erl
zuiderkwast: Actually, I think it's enough (and more clear) to use `<<?LOAD_LOCAL_FILE_REQ, FileName/binary>> ->` directly here. It's a very simple structure here. (For the others cases, e.g. `?error_pattern`, it was a more complex structure of so I preferred a macro for the pattern and a separate function to parse the packet.)

The name `LOAD_LOCAL_FILE_REQ` is good.

README.md
```diff
@@ -111,8 +111,10 @@ The encode and protocol test suites does not require a
 running MySQL server on localhost.
 
 For the suites `mysql_tests`, `ssl_tests` and `transaction_tests` you need to
-start MySQL on localhost and give privileges to the user `otptest` additionally you need to set my.cnf with `local_infile=1`. And (for
-`ssl_tests`) to the user `otptestssl`:
+start MySQL on localhost and give privileges to the user `otptest`. 
+Additionally for `mysql_tests` the mysql server (my.cnf) needs to be configured 
+to allow file loading by adding the following the mysql configuration `local_infile=1`. 
+And (for `ssl_tests`) to the user `otptestssl`:
```
zuiderkwast: It looks better, but I still don't like that this info is in the middle of how to setup test user privileges.

I think it would be better to put this about `local_infile=1` after the `GRANT` code block, next to the SSL files and config.

Is `local_infile=1` the default? The test passes in Travis without this, so it must be enabled by default in Travis at least.

If it's on by default, I think we should write something along the lines "Local infile needs to be enabled in the server, which it is by default. If you're having problems with the LOAD LOCAL INFILE test case, check that `local_infile=1` in my.cnf". But I don't know if it's on by default. What do you think?

zuiderkwast:
Hi @stuart-thackray! Sorry for posting the comments one by one. I could have done it in one review.

> I was using gitkraken and was doing line by line commits; and accidently included that swp file.

OK. I don't know gitraken but it seems to take away the control of what's happening. :-) It's fine, as long as we see things like this in the code review.

> Not sure if you use linux/vim; but you can add `% vim: sw=4 ts=4 et ft=erlang` to the top of the file

Yes, I know. There are several ways for different editors, e.g. there is some standard where you put an `.editorconfig` in the repo. I think these are all a bit cluttering the project, but a single vim-line is probably the least cluttered way. Emacs seems to be the preferred editor of Erlang people, so the Erlang Emacs mode has this style as the default. I'm using gedit myself. Anyway, 4 spaces seems to be the default for Erlang so people should use an editor where it's possible to set a default per language and to adapt the style to that of surrounding code. It's not only indentation but also other things like how variable names are chosen, etc. I don't want to imply which editor to use.

> I have pushed the changes requested; let me know if I missed any tried to go one by one.

It looks good in general! I have looked though the changes and added some more comments.

README.md
stuart-thackray: > Is `local_infile=1` the default? The test passes in Travis without this, so it must be enabled by default in Travis at least.

Try out gitkraken; it makes git a lot easier. 

No `local_infile=1` it is not default added it to the ssl configuration file; as far as test went on fedora even with sudo I couldn't append it to the mysql configure it is read and write only root; read group and user. I couldn't figure out a command to add it via the .travis file; I am not used to do either test cases for travis builds.

stuart-thackray:
Thanks; I find it interesting to hear about others experience/usage of diffrent IDEs, etc. 
Didn't know about `.editorconfig` always interesting to find out new stuff. 

Interesting that you use emacs; I use erlide a plugin to ecipse builds real time etc (but more for the library managemet); I use vi/vim mostly if pushing small amount of code to a repo to try and keep to the standards of a repo. Didn't use for those changes as was using erlied; and it doesn't use either the vim or .editorconfig but was coping and pasting I didn't write/figure all code myself; a colegue did some.

README.md
zuiderkwast: Indeed, gitkraken looks like it gives a good overview of the branches and commits. I have only looked briefly at the website.

You can skip the travis file. The test case is passing in the builds so it seems to be enabled on Travis already.

zuiderkwast: Hi! What's happening to this PR? This seems to be the only unresolved comment. Apart from that, everything seems to be done, or am I missing something?

stuart-thackray:
Closed on account of newer PR 168

Issue #94, opened 2018-11-07
Retrieve partial result sets


juhlig:
The current implementation always retrieves the full result set in a `query` or `execute` call. As I am sometimes dealing with very large result sets, I fear I may run out of memory.

For that reason, it would be nice to have a possibility to fetch only one (or some) resultset rows on demand, let the application process them, then fetch the next, and so on. Somewhat in the style of `match/{3,1}` and `select/{3,1}` in `ets`, you know?

zuiderkwast:
Yes, I'm positive to this feature. An optional `Limit` parameter to all the query and execute functions then, and a `next/1` to get the next chunk? An the return values of query and execute have to be extended with a continuation. Or perhaps we can only have one such chunked result waiting at a time. Then we don't need a continuation. Just a `next/0`.

How to combine chunking with calls that may returns multiple different result sets? Perhaps we don't need to support that...

```Erlang
{ok, [ResultFoo, ResultBar]} = mysql:query(Pid, "SELECT * foo; SELECT * from bar;").
```

Feel free to implement it!

juhlig:
I think there are a few issues with a chunked retrieval that complicate integration into the current functions.

Namely, the resultset (including the coldefs) of a query is only available to the client when the resultset of a previous query has been fetched completely.

As an illustration, if you issue a query like `SELECT * FROM foo` and don't fetch anything yet, this is what is in the receive buffer:
`[colcount] [coldef] ... [eof] [row] ... [eof]`
If you then issue another query (or use a multi statement like in your comment) like `SELECT * FROM bar`, this is what is in the receive buffer:
`[colcount] [coldef] ... [eof] [row] ... [eof] [colcount] [coldef] ... [eof] [row] ... [eof]`

By using chunked fetching, a client must operate in a more sequential fashion and obey to some additional restrictions, and/or use multiple connections. I don't see how the library could sensibly handle this, so it is in the responsibility of the client.

Specifically, it is not possible (with 1 connection) to read a resultset chunked and update something, even if totally unrelated, or fetch additional information while processing it. The C API handles this dilemma in a straightforward manner: it just forbids it; if you want it, you have to use a separate connection.

Also, multi statements are not really possible, at least not in a manner that would return something like the current `query` function does. The only way I see to achieve this would be by using two continuations, one for "more rows in a resultset", one for "more resultsets in response".

[...]

PR #104, opened 2019-01-28
result filter/map capability

juhlig:
This PR adds the capability to pass a one-ary or two-ary function along with `query` and `execute` calls, to be used somewhat like `lists:filtermap/2`, as discussed in Issue #94.

Result rows will be passed to this function during fetching. The output of the function determines if the row is included or excluded from the result (filter capability), and may at the same time transform the row into something else (mapping capability). The function may also use side effects, eg send the row to another process, write to a file, insert into `ets` tables, etc. An important point is that rows are fetched and decoded only as fast as the function handles them (may be fast if it does some simple filtering or mapping or asynchronously hands them off to another process, or slow if it needs to do complicated things or stuff that is slow by nature), so it can also work as a throttling mechanism to prevent flooding memory with data that is needed only temporarily (ie, not all at once).

To achieve this behavior, it was necessary to change the response fetching strategy (`mysql_protocol:fetch_response` etc). Before, it was like this:
* Fetch _all_ row packets of a result set
* Decode the rows
* Repeat for the next result set until there are none left in the response

Now, it has been changed to:
* Fetch 1 row packet of a result set
* Decode this row packet to a row
* Pass the decoded row to the filtermap function, and depending on the return value (mimicking `lists:filtermap/2` behavior):
  * if `true`, keep the row (to return later as the return of the `query` or `execute` call) as is
  * if `false`, drop the row
  * if `{true, Value}`, keep `Value` instead of the original row (to return later)
* Repeat for next row until there are no more row packets in the result set
* Repeat for the next result set until there are none left in the response

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -49,14 +49,15 @@
 -type row() :: [term()].
 -type rows() :: [row()].
 
+-type query_filtermap_fun() :: fun((row()) -> query_filtermap_res())
+                             | fun((column_names(), row()) -> query_filtermap_res()).
+-type query_filtermap_res() :: boolean()
+                             | {true, term()}.
+
 -type query_result() :: ok
                       | {ok, column_names(), rows()}
                       | {ok, [{column_names(), rows()}, ...]}
                       | {error, server_reason()}.
-
--type query_filtermap() :: undefined
-                         | fun((row()) -> boolean() | {true, term()})
-                         | fun((column_names(), row()) -> boolean() | {true, term()}).
 
 -define(default_connect_timeout, 5000).
 
@@ -166,101 +167,75 @@ start_link(Options) ->
     end,
     Ret.
 
-%% @doc Executes a query with the query timeout as given to start_link/1.
-%%
-%% It is possible to execute multiple semicolon-separated queries.
-%%
-%% Results are returned in the form `{ok, ColumnNames, Rows}' if there is one
-%% result set. If there are more than one result sets, they are returned in the
-%% form `{ok, [{ColumnNames, Rows}, ...]}'.
-%%
-%% For queries that don't return any rows (INSERT, UPDATE, etc.) only the atom
-%% `ok' is returned.
+%% @see query/5.
 -spec query(Conn, Query) -> Result
     when Conn :: connection(),
          Query :: iodata(),
          Result :: query_result().
 query(Conn, Query) ->
-    query_call(Conn, {query, Query, undefined, undefined}).
+    query(Conn, Query, no_params, no_filtermap_fun, default_timeout).
 
-%% @doc Depending on the 3rd argument this function does different things.
-%%
-%% If the 3rd argument is a list, it executes a parameterized query and applies
-%% no filtering/mapping of the result rows. This is equivallent to query/5
-%% with undefined as filter/map function and the query timeout as given to
-%% start_link/1.
-%%
-%% If the 3rd argumeent is a function, it is used to filter/map the result
-%% rows. This is equivalent to query/4 with the query timeout given to
-%% start_link/1.
-%%
-%% If the 3rd argument is a timeout, it executes a plain query with this
-%% timeout and applies no filtering/mapping of the resultset rows.
-%%
-%% The return value is the same as for query/2.
-%%
-%% @see query/2.
-%% @see query/4.
+%% @see query/5.
 -spec query(Conn, Query, Params | FilterMap | Timeout) -> Result
     when Conn :: connection(),
          Query :: iodata(),
-         Timeout :: timeout(),
-         Params :: [term()],
-         FilterMap :: query_filtermap(),
+         Timeout :: default_timeout | timeout(),
+         Params :: no_params | [term()],
+         FilterMap :: no_filtermap_fun | query_filtermap_fun(),
          Result :: query_result().
-query(Conn, Query, Params) when is_list(Params) ->
-    query_call(Conn, {param_query, Query, Params, undefined, undefined});
-query(Conn, Query, FilterMap) when FilterMap=:=undefined;
-        is_function(FilterMap, 1); is_function(FilterMap, 2) ->
-    query_call(Conn, {query, Query, FilterMap, undefined});
-query(Conn, Query, Timeout) when is_integer(Timeout); Timeout == infinity ->
-    query_call(Conn, {query, Query, undefined, Timeout}).
+query(Conn, Query, Params) when Params == no_params;
+                                is_list(Params) ->
+    query(Conn, Query, Params, no_filtermap_fun, default_timeout);
+query(Conn, Query, FilterMap) when FilterMap == no_filtermap_fun;
+                                   is_function(FilterMap, 1);
+                                   is_function(FilterMap, 2) ->
+    query(Conn, Query, no_params, FilterMap, default_timeout);
+query(Conn, Query, Timeout) when Timeout == default_timeout;
+                                 is_integer(Timeout);
+                                 Timeout == infinity ->
+    query(Conn, Query, no_params, no_filtermap_fun, Timeout).
 
-%% @doc Depending on the 3rd and 4th arguments this function does different
-%% things.
-%%
-%% If either the 3rd argument is a list, a prepared statement is created,
-%% executed and then cached for a certain time. If the same query is executed
-%% again when it is already cached, it does not need to be prepared again.
-%% The minimum time the prepared statement is cached can be specified using the
-%% option `{query_cache_time, Milliseconds}' to start_link/1.
-%% Conversely, if the 3rd argument is not a list, a plain query is executed.
-%%
-%% If the 4th argument is a timeout, the query is executed with that timeout.
-%% Conversely, if the 4th argument is not a timeout, the timeout given in
-%% start_link/1 is used.
-%%
-%% If either the 3rd or 4th argument is a function instead of a list or
-%% timeout, respectively, it is used to filter/map the result rows.
-%%
-%% The return value is the same as for query/2.
+%% @see query/5.
 -spec query(Conn, Query, Params, Timeout) -> Result
         when Conn :: connection(),
              Query :: iodata(),
-             Timeout :: timeout(),
-             Params :: [term()],
+             Timeout :: default_timeout | timeout(),
+             Params :: no_params | [term()],
              Result :: query_result();
     (Conn, Query, FilterMap, Timeout) -> Result
         when Conn :: connection(),
              Query :: iodata(),
-             Timeout :: timeout(),
-             FilterMap :: query_filtermap(),
+             Timeout :: default_timeout | timeout(),
+             FilterMap :: no_filtermap_fun | query_filtermap_fun(),
              Result :: query_result();
     (Conn, Query, Params, FilterMap) -> Result
         when Conn :: connection(),
              Query :: iodata(),
-             Params :: [term()],
-             FilterMap :: query_filtermap(),
+             Params :: no_params | [term()],
+             FilterMap :: no_filtermap_fun | query_filtermap_fun(),
              Result :: query_result().
-query(Conn, Query, Params, Timeout) when is_list(Params) andalso (is_integer(Timeout) orelse Timeout=:=infinity) ->
-    query_call(Conn, {param_query, Query, Params, undefined, Timeout});
-query(Conn, Query, FilterMap, Timeout) when (is_function(FilterMap, 1) orelse is_function(FilterMap, 2)) andalso (is_integer(Timeout) orelse Timeout=:=infinity) ->
-    query_call(Conn, {query, Query, FilterMap, Timeout});
-query(Conn, Query, Params, FilterMap) when is_list(Params) andalso (is_function(FilterMap, 1) orelse is_function(FilterMap, 2)) ->
-    query_call(Conn, {param_query, Query, Params, FilterMap, undefined}).
+query(Conn, Query, Params, Timeout) when (Params == no_params orelse
+                                          is_list(Params)) andalso
+                                         (Timeout == default_timeout orelse
+                                          is_integer(Timeout) orelse
+                                          Timeout == infinity) ->
+    query(Conn, Query, Params, no_filtermap_fun, Timeout);
+query(Conn, Query, FilterMap, Timeout) when (FilterMap == no_filtermap_fun orelse
+                                             is_function(FilterMap, 1) orelse
+                                             is_function(FilterMap, 2)) andalso
+                                            (Timeout == default_timeout orelse
+                                             is_integer(Timeout) orelse
+                                             Timeout=:=infinity) ->
+    query(Conn, Query, no_params, FilterMap, Timeout);
+query(Conn, Query, Params, FilterMap) when (Params == no_params orelse
+                                            is_list(Params)) andalso
+                                           (FilterMap == no_filtermap_fun orelse
+                                            is_function(FilterMap, 1) orelse
+                                            is_function(FilterMap, 2)) ->
+    query(Conn, Query, Params, FilterMap, default_timeout).
 
 %% @doc Executes a parameterized query with a timeout and applies a filter/map
-%% function to the result rows..
+%% function to the result rows.
 %%
 %% A prepared statement is created, executed and then cached for a certain
 %% time. If the same query is executed again when it is already cached, it does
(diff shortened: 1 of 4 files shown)


src/mysql.erl
```diff
@@ -175,32 +181,86 @@ start_link(Options) ->
...
+             Result :: query_result().
+query(Conn, Query, Params, Timeout) when is_list(Params) andalso (is_integer(Timeout) orelse Timeout=:=infinity) ->
+    query_call(Conn, {param_query, Query, Params, undefined, Timeout});
+query(Conn, Query, FilterMap, Timeout) when (is_function(FilterMap, 1) orelse is_function(FilterMap, 2)) andalso (is_integer(Timeout) orelse Timeout=:=infinity) ->
+    query_call(Conn, {query, Query, FilterMap, Timeout});
+query(Conn, Query, Params, FilterMap) when is_list(Params) andalso (is_function(FilterMap, 1) orelse is_function(FilterMap, 2)) ->
+    query_call(Conn, {param_query, Query, Params, FilterMap, undefined}).
+
+%% @doc Executes a parameterized query with a timeout and applies a filter/map
+%% function to the result rows..
```
zuiderkwast: Yep....

I think we should put the full descriptions here and just put "see query/5" at the other other query/2,3,4 functions.

src/mysql.erl
```diff
@@ -175,32 +181,86 @@ start_link(Options) ->
          Query :: iodata(),
          Result :: query_result().
 query(Conn, Query) ->
-    query_call(Conn, {query, Query}).
+    query_call(Conn, {query, Query, undefined, undefined}).
```
zuiderkwast: Suggesting more descriptive atom names, e.g. `no_filtermap_fun` and `default_timeout`. (See also other comments about this.)

src/mysql.erl
```diff
@@ -175,32 +181,86 @@ start_link(Options) ->
...
     when Conn :: connection(),
          Query :: iodata(),
          Timeout :: timeout(),
          Params :: [term()],
+         FilterMap :: query_filtermap(),
          Result :: query_result().
 query(Conn, Query, Params) when is_list(Params) ->
-    query_call(Conn, {param_query, Query, Params});
+    query_call(Conn, {param_query, Query, Params, undefined, undefined});
+query(Conn, Query, FilterMap) when FilterMap=:=undefined;
```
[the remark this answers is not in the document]
juhlig: I think I slightly disagree here, or at least I'm uncertain when it comes to conditionally using a function or not (the exmaples below should make clear what I'm trying to say).

If we don't allow `undefined` (or whatever else to indicate that no function is to be used, I'll stick to `undefined` for the time being), it has to be done either like this:
```erlang
Res=
case X of
    foo -> mysql:query(Conn, Query, fun ([A|_]) -> A<10 end);
    bar -> mysql:query(Conn, Query)
end
```
or like this:
```erlang
FilterMap=
case X of
    foo -> fun ([A|_]) -> A<10 end;
    bar -> fun (_) -> true end
end,
Res=mysql:query(Conn, Query, FilterMap)
```

If we allow `undefined`, this is also possible, on top of the other two approaches:
```erlang
FilterMap=
case X of
    foo -> fun ([A|_]) -> A<10 end;
    bar -> undefined
end,
Res=mysql:query(Conn, Query, FilterMap)
```

src/mysql.erl
```diff
@@ -175,32 +181,86 @@ start_link(Options) ->
...
+-spec query(Conn, Query, Params | FilterMap | Timeout) -> Result
     when Conn :: connection(),
          Query :: iodata(),
          Timeout :: timeout(),
          Params :: [term()],
+         FilterMap :: query_filtermap(),
          Result :: query_result().
 query(Conn, Query, Params) when is_list(Params) ->
-    query_call(Conn, {param_query, Query, Params});
+    query_call(Conn, {param_query, Query, Params, undefined, undefined});
```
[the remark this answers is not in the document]
juhlig: Sure :) In conjunction with the previous comment, if we allow `undefined` for the API functions (`query`, `execute`), they should be made public, however.

src/mysql.erl
```diff
@@ -175,32 +181,86 @@ start_link(Options) ->
...
+%% timeout, respectively, it is used to filter/map the result rows.
+%%
+%% The return value is the same as for query/2.
+-spec query(Conn, Query, Params, Timeout) -> Result
+        when Conn :: connection(),
+             Query :: iodata(),
+             Timeout :: timeout(),
+             Params :: [term()],
+             Result :: query_result();
+    (Conn, Query, FilterMap, Timeout) -> Result
```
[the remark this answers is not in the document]
juhlig: Thanks :)

src/mysql.erl
juhlig: Hm, for that to be consistent, it should be possible to have a `query/5` call that is equivalent to a `query/2` call (in other words, `query/2` should be able to map to `query/5`), which is in fact _not_ possible, as `query/5` is always a parameterized query, while `query/2` is always a plain query.  It is possible to have equivalent _outcomes_ by giving a plain query and an empty list to `query/5`, but it would be handled different internally. Not sure it matters to end users, but it may make a difference to people working on `mysql-otp`.

Maybe it would be clearer to say something like "for an explanation of the _arguments_, see `query/5`" or something?

src/mysql.erl
juhlig: I have no preference here, can do :) See my comment on (not) allowing `undefined`, though.

src/mysql.erl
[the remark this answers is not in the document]
juhlig: On second thought, this would be an inconsistency with the current timeout behavior, since it is not possible to specify `default_timeout` or an equivalent... Therefore, I'll disallow `undefined` for the `FilterMap` parameter as you suggested.

zuiderkwast:
I sort of agree about the documentation. I wanted to use edoc just because it's an OTP standard application. I didn't want any dependencies of any sort, apart from OTP. This project can build using `rebar` (using default values, no config), `make` or just using the OTP compile or make functions in an Erlang shell.

However, building the docs is not necessary when using the project as a dependency. Asciidoc can be nicely displayed in the github web interface as well I suppose, just like markdown, so there isn't even a need to build it. Is that right?

Anyway, let's do that separately from this change.

src/mysql.erl
zuiderkwast: That's true.

So, what if we allow `no_params`, `no_filtermap_fun` and `default_timeout` in `query/5`? Then all other variants can simply call this one.

Then, we can write in the docs "Rather than passing `no_params`, `no_filtermap_fun` and `default_timeout`, consider using `query/2,3,4` instead" as a suggestion, while still allowing fun logic like `mysql:query(Pid, Query, if X -> fun(X) -> {true, Y} end; true -> no_filtermap_fun)`.

As for the style with optional arguments, it is used extensively in OTP, e.g. `proplists:get_value/2,3`, which is fine as long as you don't want to omit some param in the middle. Originally, the only optional argument was Params, which was fine then, but it was extended later with more optional params.....

juhlig:
> However, building the docs is not necessary when using the project as a dependency. Asciidoc can be nicely displayed in the github web interface as well I suppose, just like markdown, so there isn't even a need to build it. Is that right?

Correct, see [here](https://github.com/ninenines/ranch/blob/master/doc/src/guide/book.asciidoc) for instance. Some IDEs like Eclipse can interpret it to varying degrees as well. And it can be read with any text editor, of course, similar to markdown.

> Anyway, let's do that separately from this change.

I agree, that's for later :)

src/mysql.erl
juhlig: > That's true.
> 
> So, what if we allow `no_params`, `no_filtermap_fun` and `default_timeout` in `query/5`? Then all other variants can simply call this one.

Ok, that's good :) I'll go that route. It may mean changes to other functions (like `prepare`, I guess) in order to make them work in a similar fashion, which is actually unrelated to this PR.

> As for the style with optional arguments, it is used extensively in OTP, e.g. `proplists:get_value/2,3`, which is fine as long as you don't want to omit some param in the middle. Originally, the only optional argument was Params, which was fine then, but it was extended later with more optional params.....

Yes, I know. As of lately, in my impression, they tend to avoid the "leave out stuff in the middle" variants that are possible if argument types make them discernible and are tending more to just using separate functions, or option maps (or proplists, before the advent of maps) for this purpose. I think if it ever came to a new/alternative API for `mysql-otp`, option maps (or rather proplists, since you are also targeting OTP releases that don't have them, like R16) would be the best way to do it. Something like
```erlang
-spec query2(Conn :: pid(), Query :: iodata(), Options :: [{Key :: atom(), Value :: term()}]) -> query_result().
...
-spec param_query2(Conn :: pid(), Query :: iodata(), Params :: [term()], Options :: [{Key :: atom(), Value :: term()}]) -> query_result().
...
```
where timeout, filtermap fun etc would go into the options.

However, this also is something to be discussed separately. Which reminds me, if you ever get to #erlang on freenode (IRC), ping me, I often hang out there during the day, nick is the same as my github handle ;)

src/mysql.erl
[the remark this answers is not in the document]
juhlig: I just noticed that this does not work with edoc... edoc will only generate documentation for the first clause :\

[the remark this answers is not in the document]
juhlig: Bummer... the guys on #erlang say that's somewhat normal as edoc just doesn't seem to support the full extend of what `-spec`s offer :( I opened an issue at bugs.erlang.org for this, but I don't think much will come of it anytime soon as it's not a pressing issue.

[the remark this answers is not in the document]
juhlig: As it stands, I see no alternative other than weakening the spec to:
```erlang
-spec query(Conn, Query, Params | FilterMap, Timeout | FilterMap) -> Result
when Conn :: connection(),
Query :: iodata(),
Timeout :: default_timeout \| timeout(),
Params :: no_params \| [term()],
FilterMap :: no_filtermap_fun \| query_filtermap_fun(),
Result :: query_result().
```
which I am reluctant to do, since it is in fact incorrect as it suggests you can pass in two filtermap functions.

juhlig:
With my last commit, I addressed most of the requested changes as best as I could. In a nutshell:

* The tests are still missing. I'll get to that later.
* I couldn't always keep the lines below 80, some are slightly over this limit. This is not due to deep nesting but because of some long names. Since I wasn't always sure about you line breaking, indentation, and alignment policies (didn't always find examples in the current code), I left it at that for now.
* I also renamed some things (like `query_filtermap` to `query_filtermap_fun`) to better reflect what they are for.
* Documentation for `query/{2,3,4}` was replaced with `@see query/5`, where I also added an extensive paragraph on the `FilterMap` argument.

src/mysql.erl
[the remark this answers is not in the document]
zuiderkwast: Keep the good spec. It's more important than EDoc. I would rather drop EDoc if it's not good enough, but that can be done later. (What I do like about EDoc though is that you get the functions and types from the code for free.)

Can you post the link to the bug you opened for EDoc? I don't think it can be that hard to solve... (or is it?)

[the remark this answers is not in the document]
juhlig: Oh, sure, here you go: [ERL-849](https://bugs.erlang.org/browse/ERL-849). Not sure about how complicated it might be, have only glimpsed into the edoc code. `edoc_specs` seems to be the module to go for, I guess.

zuiderkwast:
> Personally, I don't care much about line lengths and never restrict myself to them in my own projects (my thinking being, this is a 24" monitor I work with, not a gameboy screen ;)), so I usually have a hard time when working on other peoples projects. But I'll adhere to your requirements, of course.

Some benefits of short lines:

* I like to have a text editor on half of the laptop screen and and a terminal on the other half.
* Side-by-side diff, even with a large font size.
* Short lines prevents too much nested code.
* Short lines are easier for the eye to read. Compare to the columns in a newspaper.
* Readable on small screens like smartphones.

src/mysql.erl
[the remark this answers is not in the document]
juhlig: I don't see that there could be a problem. Before, it would allow an `integer` or `infinity` for timeouts, or a list for params. Now it will additionally allow `default_timeout` for timeouts, `no_params` for params, and `no_filtermap_fun` or a one-ary or two-ary function for filtermap. All things are additional, nothing has been removed.

src/mysql.erl
zuiderkwast: If we ever change this interface, we can make a version 2.0 and drop support for those old OTP versions. Maps are available from OTP 19.

I'll find you on IRC!

juhlig: IIRC, since OTP 17, even :)

test/mysql_tests.erl
```diff
@@ -279,6 +280,53 @@ basic_queries(Pid) ->
...
+    FilterMap2 = fun
+        (_, Row2) ->
+            FilterMap1(Row2)
+    end,
+
+    Expected = [{3, <<"text 3">>}, [1, <<"text 1">>]],
+
+    %% test with plain query
+    {ok, _, Rows1}=mysql:query(Pid, Query, FilterMap1),
+    ?assertEqual(Expected, lists:sort(Rows1)),
```
zuiderkwast: I suggest you add `ORDER BY id` to the SQL query instead of these calls to `lists:sort/1`. Relying on the fact that tuples < lists in erlang isn't the most obvious thing.

src/mysql.erl
```diff
@@ -269,14 +244,84 @@ query(Conn, Query, Params, FilterMap) when is_list(Params) andalso (is_function(
...
+%%
+%% The value returned is then used to decide if the row is to be included in
+%% the result(s) returned from the `query' call (filtering), or if something
+%% else is to be included in the result instead (mapping). You may also use
+%% this function for side effects, like writing rows to disk or sending them
+%% to another process etc.
+%%
+%% Here is an example showing some of the things that are possible:
+%% ```
+%% Query = "SELECT `a`, `b`, `c` FROM `foo`",
```
zuiderkwast: Very good documentation!

Only one thing: I don't like the backticks around column names in MySQL examples though (nor in actual programs) because it's not standard SQL and it's not needed. Just remove them, I'd prefer.

Regarding not having the full manual in the code, have you seen that there is an edoc file in the doc/ folder? Just saying. I don't mind that the full documentation about this is here in the code for now. We can move this to e.g. asciidocs some other time.

src/mysql.erl
```diff
@@ -175,32 +181,86 @@ start_link(Options) ->
...
-    query_call(Conn, {query, Query}).
+    query_call(Conn, {query, Query, undefined, undefined}).
 
 %% @doc Depending on the 3rd argument this function does different things.
 %%
-%% If the 3rd argument is a list, it executes a parameterized query. This is
-%% equivallent to query/4 with the query timeout as given to start_link/1.
+%% If the 3rd argument is a list, it executes a parameterized query and applies
+%% no filtering/mapping of the result rows. This is equivallent to query/5
+%% with undefined as filter/map function and the query timeout as given to
```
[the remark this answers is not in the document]
zuiderkwast: > not saying I can do it *well*, explaining things is one of my weak points, especially so since English is not my native language

That's not true :-)

test/mysql_tests.erl
juhlig: Ok, no problem :)

src/mysql.erl
juhlig: > Only one thing: I don't like the backticks around column names in MySQL examples though (nor in actual programs) because it's not standard SQL and it's not needed. Just remove them, I'd prefer.

Sure, done :)

> Regarding not having the full manual in the code, have you seen that there is an edoc file in the doc/ folder? Just saying. I don't mind that the full documentation about this is here in the code for now. We can move this to e.g. asciidocs some other time.

Oh, no, I didn't notice. Looking at it, though, it seems that it would require some more structure and context, if you know what I mean. Like, just slapping the query paragraph from `mysql.erl` in there would look half-heartedly, ie not too good ;) I feel that there should be all things from connectin through querying in all sorts of ways etcetc in it, too. That's beyond my scope for now, though, so I'd say we leave the manual in the code for the time being until we can do an all-out manual ;)

zuiderkwast: Good. I agree.

test/mysql_tests.erl
```diff
@@ -279,6 +280,53 @@ basic_queries(Pid) ->
 
     ok.
 
+filtermap_queries(Pid) ->
+    ok = mysql:query(Pid, ?create_table_t),
+    ok = mysql:query(Pid, <<"INSERT INTO t (id, tx) VALUES (1, 'text 1')">>),
+    ok = mysql:query(Pid, <<"INSERT INTO t (id, tx) VALUES (2, 'text 2')">>),
+    ok = mysql:query(Pid, <<"INSERT INTO t (id, tx) VALUES (3, 'text 3')">>),
```
zuiderkwast: Just in case you didn't know you can do this in SQL:
```SQL
INSERT INTO t (id, tx) VALUES (1, 'text 1'), (2, 'text 2'), (3, 'text 3');
```

juhlig: I know ;) That's more or less an artifact there, I wasn't sure how many values I'd need, so just duplicated the line a few times over. I can change it to the `(...), (...), (...)`style if you like that better.

zuiderkwast: No, it's not important.

zuiderkwast: Sure, I'll merge it tonight or tomorrow.

src/mysql.erl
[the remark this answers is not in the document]
juhlig: Ok, [ERL-849](https://bugs.erlang.org/browse/ERL-849) has just been resolved as "won't fix"...

PR #105, opened 2019-02-04
implement COM_CHANGE_USER

juhlig:
This PR implements the `COM_CHANGE_USER` capability which allows changing of the MySQL user on an active connection, ie without closing and re-opening the connection. This is mainly useful if `mysql-otp` is to be used in a connection pool.

This is a work in progress, so there are no tests yet and no documentation beyond the bare specs. Also, some things can certainly be made prettier, for example the duplications in the queries/prepares after connect could be moved to a separate function, and part of the typespecs of `start_link/1` could be unified with that of `change_user/4`.

If the transition is successful, the connection is forthwith owned by the new user; if it fails, it stays with the old user. However it turns out, the session state is reset both in case of success and failure, ie user variables, prepared statements, temporary tables etc will be lost. One could think of the whole process as logging out the old user, attempting login with the new user, and if it fails, re-login with the old user.

So, obviously, active transactions are implicitly rolled back. Some of the other commands in `mysql-otp` notify that this happened by returning something like `implicit_rollback` (or `implicit_commit` for that matter). I didn't do that here, I believe this should be in the documentation only.

A point that may cause confusion and should therefore be documented concerns on-connect queries and prepares. As the ones given in the initial `start_link/1` (or a previous `change_user/4`, see below) are most likely of no use, actually even rather dangerous, after changing the user, those are not run again. Instead, new ones can be given via `change_user/4`. Those are only run if the change was _successful_, but the old ones are _not run again if the change failed_, as, again, this is dangerous. But a user may set some user or session variables with an on-connect query and expect them to be still there after user change failure, which is not the case.

I refrained from offering the possibility to set a timeout for the process, since that might also be dangerous. When the `KILL` reaches the server, the user change may or may not have happened, so it would be uncertain what user the connection is under after a timeout. The switch should happen quickly in normal circumstances, if it doesn't, something else is likely wrong.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -30,7 +30,8 @@
          prepare/2, prepare/3, unprepare/2,
          warning_count/1, affected_rows/1, autocommit/1, insert_id/1,
          encode/2, in_transaction/1,
-         transaction/2, transaction/3, transaction/4]).
+         transaction/2, transaction/3, transaction/4,
+         change_user/3, change_user/4]).
 
 -export_type([connection/0, server_reason/0, query_result/0]).
 
@@ -147,22 +148,9 @@ start_link(Options) ->
     end,
     case Ret of
         {ok, Pid} ->
-            %% Initial queries
-            Queries = proplists:get_value(queries, Options, []),
-            lists:foreach(fun (Query) ->
-                              case mysql:query(Pid, Query) of
-                                  ok -> ok;
-                                  {ok, _, _} -> ok;
-                                  {ok, _} -> ok
-                              end
-                          end,
-                          Queries),
-            %% Prepare
-            Prepare = proplists:get_value(prepare, Options, []),
-            lists:foreach(fun ({Name, Stmt}) ->
-                              {ok, Name} = mysql:prepare(Pid, Name, Stmt)
-                          end,
-                          Prepare);
+            execute_after_connect(Pid,
+                                  proplists:get_value(queries, Options, []),
+                                  proplists:get_value(prepare, Options, []));
         _ -> ok
     end,
     Ret.
@@ -570,6 +558,12 @@ execute_transaction(Conn, Fun, Args, Retries) ->
             %% Returning 'atomic' or 'aborted' would both be wrong. Raise an
             %% exception is the best we can do.
             erlang:raise(error, E, ?GET_STACK(Stacktrace));
+        ?EXCEPTION(error, change_user_in_transaction = E, Stacktrace) ->
+            %% The called tried to change user inside the transaction, which
+            %% is not allowed and a serious mistake. We roll back and raise
+            %% an error.
+            ok = gen_server:call(Conn, rollback, infinity),
+            erlang:raise(error, E, ?GET_STACK(Stacktrace));
         ?EXCEPTION(Class, Reason, Stacktrace) ->
             %% We must be able to rollback. Otherwise let's crash.
             ok = gen_server:call(Conn, rollback, infinity),
@@ -580,6 +574,62 @@ execute_transaction(Conn, Fun, Args, Retries) ->
                 exit  -> Reason
             end,
             {aborted, Aborted}
+    end.
+
+%% @doc Equivalent to `change_user(Conn, Username, Password, [])'.
+%% @see change_user/4
+-spec change_user(Conn, Username, Password) -> Result
+    when Conn :: connection(),
+         Username :: iodata(),
+         Password :: iodata(),
+         Result :: ok.
+change_user(Conn, Username, Password) ->
+    change_user(Conn, Username, Password, []).
+
+%% @doc Changes the user of the active connection without closing and
+%% and re-opening it. The currently active session will be reset (ie,
+%% user variables, temporary tables, prepared statements, etc will
+%% be lost) independent of whether the operation succeeds or fails.
+%%
+%% If change user is called when a transaction is active (ie, neither
+%% committed nor rolled back), calling `change_user' will fail with
+%% an error exception and `change_user_in_transaction' as the error
+%% message.
+%%
+%% If the change user operation fails for other reasons (eg authentication
+%% failure), an error exception occurs, and the connection process
+%% exits with reason `change_user_failed'. The connection can not be used
+%% any longer if this happens.
+%%
+%% For a description of the `database', `queries' and `prepare'
+%% options, see `start_link/1'.
+%%
+%% @see start_link/1
+-spec change_user(Conn, Username, Password, Options) -> Result
+    when Conn :: connection(),
+         Username :: iodata(),
+         Password :: iodata(),
+         Options :: [Option],
+         Result :: ok,
+         Option :: {database, iodata()}
+                 | {queries, [iodata()]}
+                 | {prepare, [NamedStatement]},
+         NamedStatement :: {StatementName :: atom(), Statement :: iodata()}.
+change_user(Conn, Username, Password, Options) ->
+    case in_transaction(Conn) of
+        true -> error(change_user_in_transaction);
+        false -> ok
+    end,
+    Database = proplists:get_value(database, Options, undefined),
+    Ret = gen_server:call(Conn, {change_user, Username, Password, Database}),
+    case Ret of
+        ok ->
+            execute_after_connect(Conn,
+                                  proplists:get_value(queries, Options, []),
+                                  proplists:get_value(prepare, Options, [])),
+            ok;
+        {error, Reason} ->
+            error(Reason)
     end.
 
 %% @doc Encodes a term as a MySQL literal so that it can be used to inside a
@@ -601,6 +651,25 @@ encode(Conn, Term) ->
     mysql_encode:encode(Term1).
 
 %% --- Helpers ---
+
+%% @doc Executes the given queries and prepares the given statements after a
+%% connection has been made.
+-spec execute_after_connect(connection(), [iodata()], [{atom(), iodata()}])
+    -> ok.
+execute_after_connect(Conn, Queries, Prepares) ->
+    lists:foreach(fun (Query) ->
+                      case query(Conn, Query) of
+                          ok -> ok;
+                          {ok, _} -> ok;
+                          {ok, _, _} -> ok
+                      end
+                  end,
+                  Queries),
+    lists:foreach(fun ({Name, Stmt}) ->
+                      {ok, Name} = prepare(Conn, Name, Stmt)
+                  end,
+                  Prepares),
+    ok.
 
 %% @doc Makes a gen_server call for a query (plain, parametrized or prepared),
 %% checks the reply and sometimes throws an exception when we need to jump out
--- a/src/mysql_protocol.erl
+++ b/src/mysql_protocol.erl
@@ -27,7 +27,7 @@
 %% @private
 -module(mysql_protocol).
 
--export([handshake/7, quit/2, ping/2,
+-export([handshake/7, change_user/6, quit/2, ping/2,
          query/4, query/5, fetch_query_response/3,
          fetch_query_response/4, prepare/3, unprepare/3,
          execute/5, execute/6, fetch_execute_response/3,
@@ -227,6 +227,27 @@ fetch_execute_response(SockModule, Socket, Timeout) ->
 
 fetch_execute_response(SockModule, Socket, FilterMap, Timeout) ->
     fetch_response(SockModule, Socket, Timeout, binary, FilterMap, []).
+
+%% @doc Changes the user of the connection.
+-spec change_user(atom(), term(), iodata(), iodata(), binary(),
+                  undefined | iodata()) -> #ok{} | #error{}.
+change_user(SockMod, Socket, Username, Password, Salt, Database) ->
+    DbBin = case Database of
+        undefined -> <<>>;
+        _ -> iolist_to_binary(Database)
+    end,
+    Hash = hash_password(Password, Salt),
+    Req = <<?COM_CHANGE_USER, (iolist_to_binary(Username))/binary, 0,
+            (lenenc_str_encode(Hash))/binary,
+            DbBin/binary, 0, ?UTF8:16/little>>,
+    {ok, _SeqNum1} = send_packet(SockMod, Socket, Req, 0),
+    {ok, Packet, _SeqNum2} = recv_packet(SockMod, Socket, infinity, any),
+    case Packet of
+        ?ok_pattern ->
+            parse_ok_packet(Packet);
+        ?error_pattern ->
+            parse_error_packet(Packet)
+    end.
 
 %% --- internal ---
 
@@ -1202,6 +1223,13 @@ lenenc_str(Bin) ->
     {Length, Rest} = lenenc_int(Bin),
     <<String:Length/binary, Rest1/binary>> = Rest,
     {String, Rest1}.
+
+%% Length-encoded-string encode. Prefixes the value with a
+%% length-encoded-integer denoting its size.
+-spec lenenc_str_encode(Input :: binary()) -> binary().
+lenenc_str_encode(Bin) ->
+    Length = byte_size(Bin),
+    <<(lenenc_int_encode(Length))/binary, Bin:Length/binary>>.
 
 %% nts/1 decodes a nul-terminated string
 -spec nulterm_str(Input :: binary()) -> {String :: binary(), Rest :: binary()}.
--- a/test/mysql_change_user_tests.erl
+++ b/test/mysql_change_user_tests.erl
@@ -1,0 +1,153 @@
+%% MySQL/OTP – MySQL client library for Erlang/OTP
+%% Copyright (C) 2014 Viktor Söderqvist
+%%
+%% This file is part of MySQL/OTP.
+%%
+%% MySQL/OTP is free software: you can redistribute it and/or modify it under
+%% the terms of the GNU Lesser General Public License as published by the Free
+%% Software Foundation, either version 3 of the License, or (at your option)
+%% any later version.
+%%
+%% This program is distributed in the hope that it will be useful, but WITHOUT
+%% ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+%% FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+%% more details.
+%%
+%% You should have received a copy of the GNU Lesser General Public License
+%% along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+%% @doc This module performs test to an actual database.
+-module(mysql_change_user_tests).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-define(user1,     "otptest").
+-define(password1, "otptest").
+-define(user2,     "otptest2").
+-define(password2, "otptest2").
+
+%% Ensure that the current user can be changed to another user
+%% when given correct credentials.
+correct_credentials_test() ->
+    Pid = connect_db(?user1),
+    ?assertEqual(ok, mysql:change_user(Pid, ?user2, ?password2)),
+    ?assert(is_current_user(Pid, ?user2)),
+    close_conn(Pid),
+    ok.
+
+%% Ensure that change user fails when given incorrect credentials,
+%% and that the current user still works.
+incorrect_credentials_fail_test() ->
+    Pid = connect_db(?user1),
+    TrapExit = erlang:process_flag(trap_exit, true),
+    ?assertError({1045, <<"28000">>, <<"Access denied", _/binary>>},
+                 mysql:change_user(Pid, ?user2, ?password1)),
+    ExitReason = receive {'EXIT', Pid, Reason} -> Reason after 1000 -> error(timeout) end,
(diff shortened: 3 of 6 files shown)


src/mysql_protocol.erl
```diff
@@ -203,6 +203,26 @@ execute(#prepared{statement_id = Id, param_count = ParamCount}, ParamValues,
 fetch_execute_response(SockModule, Socket, Timeout) ->
     fetch_response(SockModule, Socket, Timeout, binary, []).
 
+-spec change_user(atom(), term(), iodata(), iodata(), binary(), iodata()) ->
+    #ok{} | #error{}.
+change_user(SockMod, Socket, Username, Password, Salt, Database) ->
```
zuiderkwast: Database can also be undefined (in the spec).

src/mysql.erl
```diff
@@ -437,6 +438,49 @@ execute_transaction(Conn, Fun, Args, Retries) ->
...
+change_user(Conn, Username, Password, Options) ->
+    Database = proplists:get_value(database, Options, undefined),
+    Ret = gen_server:call(Conn, {change_user, Username, Password, Database}),
+    case Ret of
+        ok ->
+            Queries = proplists:get_value(queries, Options, []),
+            lists:foreach(fun (Query) ->
+                              case query(Conn, Query) of
+                                  ok -> ok;
+                                  {ok, _} -> ok;
```
zuiderkwast: Some duplicated code from handshake code here? Please move shared code out to a function.

src/mysql.erl
```diff
@@ -437,6 +438,49 @@ execute_transaction(Conn, Fun, Args, Retries) ->
             {aborted, Aborted}
     end.
 
+-spec change_user(Conn, Username, Password) -> Result
+    when Conn :: connection(),
+         Username :: iodata(),
+         Password :: iodata(),
+         Result :: ok | {error, server_reason}.
```
zuiderkwast: `server_reason()`

src/mysql.erl
```diff
@@ -437,6 +438,49 @@ execute_transaction(Conn, Fun, Args, Retries) ->
...
+change_user(Conn, Username, Password) ->
+    gen_server:call(Conn, {change_user, Username, Password, undefined}).
+
+-spec change_user(Conn, Username, Password, Options) -> Result
+    when Conn :: connection(),
+         Username :: iodata(),
+         Password :: iodata(),
+         Options :: [Option],
+         Result :: ok | {error, server_reason},
```
zuiderkwast: `server_reason()`

zuiderkwast:
As for the on-connect queries, implicit commit (I believe it's implicit commit, not rollback?), etc. just test and document the behaviour and we're fine.

Just an idea: If changing user fails and there are any on-connect queries or prepares defined, I was thinking that perhaps it would be more consistent to crash, not to leave the connection in a surprising state.

src/mysql.erl
juhlig: Yes, an oversight on my part, good find.

src/mysql.erl
juhlig: Copied over from the previous section, will fix.

src/mysql.erl
juhlig: Yes, I just wanted to make a rough sketch of it first to show you before I make it pretty ;) That will be straightened out in the final version, of course.

src/mysql_protocol.erl
juhlig: Ah, yes, right. Will fix.

juhlig:
As for the mysql 8 PR, I guess this will have interferences, yes. I will look at their PR later, maybe it can be mitigated with an intermediate step from which both can proceed independently again.

juhlig:
On on-connect queries/prepares and implicit commit/rollback of transactions, those are actually different topics, but the effect of a user change affects both. To illustrate, consider the following scenariio:

```
user connects:
  {ok, P}=mysql:start_link([{user, "foo"}, {password, "foo"}, {queries, ["SET @foovar='abc'"]}, {prepare, [{foostmt, "SELECT a FROM test where b=?"}]}]).

-> variable foovar is set to "abc"
-> statement foostmt is prepared

user creates a temporary table:
  mysql:query(P, "CREATE TEMPORARY TABLE bar (x INT)")

-> temporary table is created and can be used

user starts transaction and issues some statements in it:
  mysql:query(P, "BEGIN"),
  mysql:query(P, "INSERT INTO test (a, b) VALUES (1, 234)")

-> transaction is still active, ie neither committed nor rolled back

user issues a change_user command:
  mysql:change_user(P, "bar", "bar", [{queries, ["SET @barvar='def'"]}, {prepare, [{barstmt, "SELECT b FROM foo WHERE a=?"}]}])
```

At this point, the user `foo` is logged out and the session is being reset, meaning:
* the user variable `foovar` is lost
* the prepared statement `foostmt` is lost
* the temporary table `bar` is lost
* the transaction is dropped, like implicitly rolled back, not committed

Now, if the `change user` operation is successful, all proceeds like expected, as if the connection had been closed and re-opened for user `bar`, subsequently executing the on-connect query setting user variable `barvar` to `"def"` and preparing the statement `barstmt`.

The tricky/confusing part is when the `change_user` operation fails (wrong password, for example). The user falls back to what it was, `foo`, but all the session-bound stuff he did is gone, as if he had disconnected and re-connected with the same user (but without the on-connect queries/prepares), namely:
* the user variable `foovar` that was there before is gone
* the prepared statement `foostmt` that was there before is gone
* the temporaray table `bar` that was there before is gone
* the transaction that was active before is gone

There are two points of confusion here:
* A user may expect that he is still in the old session with all the transient things he did there when a `change_user` fails, instead of in a new (reset) session. As this is really a MySQL thing that is documented there and that we can do nothing about, this may actually be ok, but...
* Even assuming knowledge of the former point, a user may expect to find the new session in a state as after the initila `start_link`, with the on-connect queries executed and the prepares made. This is not the case, however, and while it would be safe to re-execute the prepares, blindly re-executing the queries is likely to be dangerous, as there could be more going on than just setting some variables (eg, all kinds of writing operations like `INSERT`, `UPDATE`, DDL statements, etc)

<hr />

Summing it all up, I would suggest to document the loss of an active transaction, and invariably crash on `change_user` failure, as this would be much more consistent.

zuiderkwast:
> Summing it all up, I would suggest to document the loss of an active transaction, and invariably crash on `change_user` failure, as this would be much more consistent.

Sounds good!

>    * the transaction is dropped, like implicitly rolled back, not committed

OK, makes sense. If would be interesting to see what happens if change user is called inside a managed transaction, i.e. a fun passed to `mysql:transaction`.

juhlig:
That's certainly a very big _don't do that_ ;) What would happen is that the part of the transaction that was done before the change user would be lost. The rest of the transaction would be done in a new session, only it really wouldn't be in a transaction any more: rollbacks would not be possible (but would return `ok` nevertheless, just tried: issuing `ROLLBACK` outside of a transaction is not an error in mysql, nor is `COMMIT`; not even a warning), and commits would be pointless; and on top of that, whatever the transaction was supposed to do, it would not be done atomically.

Scenario:
```
<begin managed transaction>
1) INSERT INTO foo (a) VALUES (1)
<change user>
2) INSERT INTO foo (a) VALUES (2)
...
3) INSERT INTO foo (a) VALUES (3)
<end managed transaction>
```
The insert at (1) would be gone completely, and the inserts at (2) and (3) would not be executed atomically (or isolated, for that matter).

src/mysql.erl
juhlig: Code duplication has been removed with my latest commit, moved into `execute_after_connect/3` (not a very imaginative name, I admit.

juhlig:
With my latest commit, I introduced a security check into change user: It first checks if a transaction is active, and if so, throws an error exception. There is also a way to explicitly bypass this check with the option `force_if_in_transaction`: if set to `true`, the check is skipped; default is `false`. The need to explicitly set this option, together with the purposely unwieldy option name, should be enough to prevent users from using it without thinking twice, without completely blocking the way to do so if absolutely desired.

Also, a failure of the change user operation will now throw an error exception, as discussed before. This can be catched and the connection will still be usable then, but it will be a virgin session under the old user then (this is also clarified in the documentation). I have been thinking of implicitly closing the connection, but I'm not sure. What do you think?

As to using exceptions instead of error tuples, this is to prevent the errors from going unnoticed, as change user might be called without checking the return value.

src/mysql_protocol.erl
juhlig: Fixed.

juhlig:
I have been playing with some tests for this, and force-changing user when in a managed transaction seems a rather baaad idea =^^= It should be disallowed when in a managed transaction altogether, though I would allow it when in an un-managed (`BEGIN...`) one, for the reason that, when used in a connection pool, the connection might be returned to the pool with an unmanaged transaction still open.

zuiderkwast:
Your suggestions seem fine. I prefer not adding any new configuration though, unless really necessary. Better just choose a good behavior and describe it.

I'm curious about your use case. Are you planning to let your users have their own database user and also their own database, like in a web hosting environment?

juhlig:
Ok, this is going to be lengthy, so better grab some snacks first ;)

> I'm curious about your use case. Are you planning to let your users have their own database user and also their own database, like in a web hosting environment?

Planning, no, we (the company I work for) already do that ;) In a nutshell, our application has a pool of accounts (read, customers) which are backed by databases on a (set of) MySQL database machines. Each account is to be restricted to its own (logical) database, and thus has its own user. The accounts are organized in a 3-tier hierarchy by the application: super (system-wide stuff), customer groups, and end-users (which are members of groups). For some things, we need data from all three tiers.

Now, the Erlang app in question acts as a kind of "smart" cache between the MySQL database and the front-end application (which is mainly PHP). Earlier (before said Erlang app), we would pull all the data (namely, sets of email addresses and address patterns) into the PHP processes and process them there. But beyond a certain size, this is rather inefficient. Data from the super tier seldom changes, but is needed often (actually, all the time), so pulling it down again and again only to be thrown away again when the PHP process terminates is inefficient. Data on the lower tiers will change more often and be used less frequently, but still.

Thus, this Erlang application came to life. The PHP processes talk to it via a TCP interface, it will pull the required data from the MySQL database into ETS tables, remember and ocassionally try to update it for as long as its being accessed in a given time frame, and drop it if its not being used for a while; it will also only fetch and contain the data once, instead of in every PHP process as before. This is the "cache" side of it. It will also do most of the processing which previously was done inside the PHP processes, so those only ask simple "questions", and the Erlang app will do all the normalization, searching etc (in parallel, so many connected PHP processes may transparently access the same data, like that from the super tier), and just give a simple answer to the "question". This is the "smart" side of it.

This Erlang app has become rather old, maybe 5 or 6 years. It's the first serious thing I ever did with Erlang, and I'm proud to say it is one of the most reliable parts of the entire system ;) Nevertheless, it could use a make-over, and I would like to use `mysql-otp` for the MySQL side of it (currently, it uses a MySQL driver I wrote myself, just the bare necessities for the greater app, since there was nothing really to my liking available back then).

What now pointed me to `change user` is that a new... ummm... "feature"... requires that a PHP process on the same machine iterates through all the accounts we have, some of them multiple times; the one-user-per-account doctrine requires it to create a new connection for each account (and the PHP MySQL driver has no change user capability, either); for most of the accounts, there is little for it to do, so it quickly closes the connection and opens the next. The effect is that it thereby leaves a wake of connections in TCP_WAIT state, so much that, together with some other processes running at the same time, it simply exhausts the available ephemereal ports of the machine.

Now, if the Erlang cache was accessed at that time and it needs to fetch some data from the MySQL databases, it simply can't connect for want of local ports. And without change user capability (which my old MySQL implementation also lacks), keeping a few connections to MySQL open/pooled is pointless, since the chances they are connected with the required user is practically nil, and I cannot switch them over to the user I need.

To be sure, it would be easy to just slap another interface into the machine, or move the entire Erlang app over to another machine. Probably, all my coworkers would do just that without a second thought. But I don't like the approach of throwing hardware at a problem until it goes away ;)

zuiderkwast:
Thanks for the explanation! This driver too was born when porting parts of a PHP system to Erlang something like 5 years ago, specifically the web API parts. I wasn't happy with the existing drivers either. :-)

Please finish this *change user* feature soon! I'm starting to like it and I want to merge it. 👍

juhlig:
How weird... my local MySQL installation (5.6.65) leaves the connection intact when change user is attempted with wrong credentials, and just falls back to the current user. The installation at travis (5.6.33) closes the connection.

MySQLs documentation on the command is sparse and inconclusive in that regard.

I will make it so the connection is always closed from the client side if change user fails, even if the MySQL version in question would leave it intact. This is IMO the cleanest solution.

juhlig:
Just pushed what will be, give or take, the final version.

I removed the option to force change user when in a transaction, managed or not, and will always generate an error exception with the message `change_user_in_transaction` (wording up for discussion ;)). This is clearer, and we can always add it later if it is ever needed.

When change user fails for other reasons (eg, wrong credentials), the connection process wil now exit with reason `change_user_failed`. I'm not totally sure about this, though, as this will cause an exit signal to the linked processes (usually, the process using the connection), and if they're not trapping exits, they will exit as well. But I think this is a clean solution.

The tests for change user are in a separate module, as they are similar but rather distinct from all the other things.

Also, the tests now require a second user to run (lacking imagination, I just named it `otptest2`). This is set up accordingly in `.travis.yml` and mentioned in `README.md`

zuiderkwast:
It looks great! Whenever you're ready.

juhlig:
I think I am :) removed the `WIP` from the title.

It will still probably impact the mysql8 PR, though...

zuiderkwast:
> It will still probably impact the mysql8 PR, though...

Yes. That one seems to be stalled anyway, which is a pity, but I won't let it block this PR. There will be a merge conflict because of the stored auth_plugin_data, but hopefully not a difficult one.

test/mysql_change_user_tests.erl
```diff
@@ -0,0 +1,153 @@
+%% MySQL/OTP – MySQL client library for Erlang/OTP
+%% Copyright (C) 2014 Viktor Söderqvist
```
zuiderkwast: Name and year are slightly incorrect :-)

juhlig: I was assuming the copyright year and author name refers to MySQL/OTP itself, not the module in question. I can change it if you want, of course.

zuiderkwast: Normally these things are per file.

The author of each piece of the code owns the copyright, so authors who add code (excluding minor changes) should be added here. I don't mind having my name here, but it's in your interest to add you name here. If I own all the code, hypothetically, I can re-license it however I want, which may not be in your interest. (I don't know how important these lines are though, because if you really want to know exactly who did what and when, you use git blame.)

juhlig: Ok, changed it :)

zuiderkwast:
# 🎈

juhlig:
yay :)

juhlig:
I'm starting to use `mysql-otp` with `poolboy` right now, employing this new feature. Works like a charm :) Conveniently, `poolboy`s worker start callback matches `mysql:start_link`, though of course it doesn't name `poolboy_worker` as a behavior.

Only drawback is that you have to remember to manually "reset" the connection to the initial user (which has no permissions and is only there to be connected). This is no problem in my setup, and I could go on just as well with any user, but I think `poolboy` could use something like a hook on checkin that could automatically do this. Maybe I'll look into that as well ;)

zuiderkwast:
Awesome! If you're making some sort of wrapper or glue for poolboy which could useful for others, we can publish it under this mysql-otp github "organisation" if you like.

juhlig:
As to `poolboy`s current abilities and restrictions, there is no glue needed ;) The `poolboy_worker` behavior defines only `start_link/1`, which incidentially matches `mysql:start_link/1` :)

zuiderkwast:
Yes, I know. We had this in mind from the beginning. :-) We have also since the beginning had this proof-of-concept project https://github.com/mysql-otp/mysql-otp-poolboy

I was thinking about the logic for checking out a connection as a specific user using credentials. But that's you application code, I guess.

juhlig:
> Yes, I know. We had this in mind from the beginning. :-)

Ah, nice :)

> We have also since the beginning had this proof-of-concept project https://github.com/mysql-otp/mysql-otp-poolboy
> 
> I was thinking about the logic for checking out a connection as a specific user using credentials. But that's you application code, I guess.

Well, what I'm doing now is, I provide poolboy with the worker (connection) start args, where I give it the credentials of that connect-only user. When I want to perform something (as another user with more permissions), I do it like:
```erlang
poolboy:transaction(
    conn_pool,
    fun (Conn) ->
        ok=mysql:change_user(Conn, PrivilegedUser, PrivilegedPassword),
        ...
        ok=mysql:change_user(Conn, UnprivilegedUser, UnprivilegedPassword)
    end
)
```

The first `change_user` is ok I think, only the process knows the appropriate credentials of the privileged user, anyway. The second, however, is ugly, as the process needs to know the credentials of the unprivileged user, also, plus this may easily been forgotten, or an error between the first and second call may cause this to never being reached. The connection will be returned to the pool as the privileged user then, with all it's last session state even, and none the wiser. More so, if you forget the first `change_user` in another place, you may then run stuff you don't really want to run in that old session.

What I'm imagining instead (but poolboy doesn't support as of now) is an optional pool argument with a function that takes the worker pid, and performs something with it, like
```erlang
poolboy:start_link(
    [
        {name, {local, foo}},
        {size, 10},
        {checkin_hook, fun (Pid) -> ok=mysql:change_user(Pid, "dummy", "dummy") end}
    ],
    WorkerArgs
```
Thus, when a function is returned to the pool, it will be transparently set back to the dummy user. Users can't forget it. If they even forget the first change_user to get at the user they actually want, they will notice early on as that user can't (or, at least, shouldn't be able to) do anything.

A wrapper module could maybe do that, also, but first thing it would need to be an own process that can carry around the unprivileged user credentials, not an API wrapper which `mysql-otp-poolboy` is. I'd rather have that in `poolboy` itself.

juhlig:
Just noticed another quirk (bug?) in `poolboy` that I might need to address to make it useful in my project...

You remember my description of the smart cache I'll be using this in? It is busy during the day when customers work, and quite idle at night when only scheduled stuff runs.

So I do need a few residual connections, to use when that said other nightly process eats all the ports, but not many. Some delays are ok during the night as well. But I will need more during the day, otherwise I'm causing delays when they are not ok.

Now, poolboy offers overflow workers for this, which are started when the residual workers are all in use (likely during the day), but not kept running (so during the night only the residual workers will be there).

Thing is, when an overflow worker fails to start, that will crash the entire pool, so in the as yet unlikely event that during that port exhaustion period the cache needs an overflow connection and can't get it, the pool will crash, and the precious residual connections are also gone.

The only way I could fence against this is by allowing no overflow workers, which in turn means I either go with a small pool and delays during the day, or a large pool that is idle throughout the night.

zuiderkwast:
> when an overflow worker fails to start, that will crash the entire pool

Oh, that's really bad! There are other worker pools. I don't know which ones are good though, but I think the growing and shrinking logic is perhaps the main difference between pool managers. Have you looked at seth/pooler, inaka/worker_pool or botsunit/poolgirl?

zuiderkwast:
To make sure the last change user is done, how about try-after?

```Erlang
do_stuff_as_user(DoStuffFun, User, Pass, UnprivilegedUser, UnprivilegedPassword) ->
    poolboy:transaction(
        conn_pool,
        fun (Conn) ->
            ok = mysql:change_user(Conn, PrivilegedUser, PrivilegedPassword),
            try
                DoStuffFun()
            after
                ok = mysql:change_user(Conn, UnprivilegedUser, UnprivilegedPassword)
            end
        end
    ).
```

I don't really see why this is necessary though. If all the accesses to the pool are though the wrapper which calls change_user as the first thing, why do you even need to change user back to unprivileged?

juhlig:
> To make sure the last change user is done, how about try-after?

Ah, good idea :)

> I don't really see why this is necessary though. If all the accesses to the pool are though the wrapper which calls change_user as the first thing, why do you even need to change user back to unprivileged?

Yeah, but I'd like to have that automated ;) So even if you forget to do it, no other processes may corrupt anything because it is in the wrong session. It's of no consequence to my current use case, though.

juhlig:
I have my hands in `poolboy` now and have submitted a PR that should avoid the pool crash on overflow worker start failure, and instead cause an exception in the client, where it belongs ;) Wonder why this apparently never bothered anyone before, the project is old and has even been used in riak, AFAIK.

A second PR should make it possible to mitigate a similar problem: when a residual worker crashes and fails to re-start, or if the pool itself restarts and one of the required number of residual workers fails to start, the pool will crash, repeatedly if housed under a supervisor in your application, finally taking it down if the restart frequency is reached. The first PR leaves that as is, but the second PR introduces the ability to have a "lazy" pool, ie it does not preemptively start or restart residual workers but starts them when requested via `checkout` or `transaction`, and then keeps them running after they have been returned to the pool, up to the specified pool size.

I hope they get reviewed soon, the project seems to be quite slow...

PR #110, opened 2019-03-13
make it compatible with OTP 21.0

GeraldXv:
Add a new function to avoid the bug caused by delayed sub binary creation optimization. This bug was found in OTP 21.0 and was fixed in OTP21.1. Ref: https://github.com/erlang/otp/pull/1911

### DIFF
--- a/mysql-otp/README.md
+++ b/mysql-otp/README.md
@@ -1,0 +1,176 @@
+MySQL/OTP
+=========
+
+[![Build Status](https://travis-ci.org/mysql-otp/mysql-otp.svg?branch=master)](https://travis-ci.org/mysql-otp/mysql-otp)
+ :link: [Test coverage (EUnit)](//mysql-otp.github.io/mysql-otp/eunit.html)
+ :link: [API documentation (EDoc)](//mysql-otp.github.io/mysql-otp/index.html)
+
+MySQL/OTP is a driver for connecting Erlang/OTP applications to MySQL and
+MariaDB databases. It is a native implementation of the MySQL protocol in
+Erlang.
+
+Some of the features:
+
+* Mnesia style transactions:
+  * Nested transactions are implemented using SQL savepoints.
+  * Transactions are automatically retried when deadlocks are detected.
+* Each connection is a gen_server, which makes it compatible with Poolboy (for
+  connection pooling) and ordinary OTP supervisors.
+* No records in the public API.
+* SSL.
+* Parametrized queries using cached prepared statements
+  ([What?](https://github.com/mysql-otp/mysql-otp/wiki/Parametrized-queries-using-cached-prepared-statements))
+* Slow queries are interrupted without killing the connection (MySQL version
+  ≥ 5.0.0)
+* Implements both protocols: the binary protocol for prepared statements and
+  the text protocol for plain queries.
+
+Requirements:
+
+* Erlang/OTP version R16B or later
+* MySQL database version 4.1 or later or MariaDB
+* No other dependencies
+* Authentication method `caching_sha2_password` is not supported. This is the
+  default in MySQL 8.0.4 and later, so you need to add
+  `default_authentication_plugin=mysql_native_password` under `[mysqld]` in e.g.
+  `/etc/mysql/my.cnf`.
+
+Synopsis
+--------
+
+```Erlang
+%% Connect (ssl is optional)
+{ok, Pid} = mysql:start_link([{host, "localhost"}, {user, "foo"},
+                              {password, "hello"}, {database, "test"},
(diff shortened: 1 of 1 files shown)


src/mysql_protocol.erl
```diff
@@ -846,6 +849,12 @@ flooring(Value) ->
         Trunc > Value -> Trunc - 1 %% for negative values
     end.
 
+%% @doc This function is designed to avoid the bug caused by delayed sub binary
+%% creation optimization. This bug was found in OTP 21.0 and was fixed in OTP
+%% 21.1. Ref: https://github.com/erlang/otp/pull/1911
+fetch_length_and_data(<<Length, Data/binary>>) ->
+    {Length, Data}.
```
zuiderkwast: I don't like the name 'fetch'. Fetch means get from some storage or network. We're just treating the argument to the function. How about `parse_length_prefixed_data`?

src/mysql_protocol.erl
```diff
@@ -786,37 +786,40 @@ decode_binary(#col{type = ?TYPE_BIT, length = Length}, Data) ->
...
     {decode_bitstring(Binary, Length), Rest};
-decode_binary(#col{type = ?TYPE_DATE}, <<Length, Data/binary>>) ->
+decode_binary(#col{type = ?TYPE_DATE}, Data) ->
     %% Coded in the same way as DATETIME and TIMESTAMP below, but returned in
     %% a simple triple.
-    case {Length, Data} of
-        {0, _} -> {{0, 0, 0}, Data};
+    {Length, Data1} = fetch_length_and_data(Data),
+    case {Length, Data1} of
+        {0, _} -> {{0, 0, 0}, Data1};
```
zuiderkwast: Can you put the function call in the case expression? Like this:

```Erlang
case fetch_length_and_data(Data) of
    {0, Rest} -> {{0, 0, 0}, Rest};
    {4, <<Y:16/little, M, D, Rest/binary>>} -> {{Y, M, D}, Rest}
end;
```
```

File: src/mysql_protocol.erl [hunk 3]
```diff
@@ -786,37 +786,40 @@ decode_binary(#col{type = ?TYPE_BIT, length = Length}, Data) ->
...
         {4, <<Y:16/little, M, D, Rest/binary>>} -> {{Y, M, D}, Rest}
     end;
-decode_binary(#col{type = T}, <<Length, Data/binary>>)
+decode_binary(#col{type = T}, Data)
   when T == ?TYPE_DATETIME; T == ?TYPE_TIMESTAMP ->
     %% length (1) -- number of bytes following (valid values: 0, 4, 7, 11)
-    case {Length, Data} of
+    {Length, Data1} = fetch_length_and_data(Data),
+    case {Length, Data1} of
```
zuiderkwast: Can you put the function call in the case expression? See previous comment.

src/mysql_protocol.erl
```diff
@@ -786,37 +786,40 @@ decode_binary(#col{type = ?TYPE_BIT, length = Length}, Data) ->
...
     %% length (1) -- number of bytes following (valid values: 0, 8, 12)
     %% is_negative (1) -- (1 if minus, 0 for plus)
     %% days (4) -- days
     %% hours (1) -- hours
     %% minutes (1) -- minutes
     %% seconds (1) -- seconds
     %% micro_seconds (4) -- micro-seconds
-    case {Length, Data} of
+    {Length, Data1} = fetch_length_and_data(Data),
+    case {Length, Data1} of
```
zuiderkwast: Can you put the function call in the case expression? See previous comment.

src/mysql_protocol.erl
zuiderkwast: The existing function `lenenc_int/1` does this. It can be used instead of this function.

The documentation you added here can be put in the commit message. I think it's enough, as long as we have automatic tests for 21.0.

GeraldXv:
I just finished the modification. There is a merge operation `d18f8a6` in the middle to synchronize with the upstream. I don't know if it has any effect.

README.md
```diff
@@ -27,8 +27,6 @@ Some of the features:
 
 Requirements:
 
-* Erlang/OTP version R16B or later, excluding 21.0 which has a bug in binary
```
zuiderkwast: I'd like to keep just "Erlang/OTP version R16B or later".

GeraldXv:
As there are some issues, I just create a new request: https://github.com/mysql-otp/mysql-otp/pull/112.
Pls check it.

zuiderkwast:
Aha, I was going to squash the commits and edit the commit message if needed, so no need to make a new PR... Anyway, I'll look at that one instead. :-)

PR #111, opened 2019-03-13
clean connection stop

juhlig:
Up to now, the way to close a `mysql-otp` connection was via `exit/2`. This has always bothered me TBH, there should be an API function, like `mysql:stop`, for this.

OTP 18 introduced [`gen_server:stop/{1,3}`](http://erlang.org/doc/man/gen_server.html#stop-1), which can be used to stop a `gen_server` in a clean manner, a nice fit for this. As `mysql-otp` wants to support older OTP versions as well, however, this PR provides a workaround in case `gen_server:stop/3` is not available, mimicking it's behavior (but is not quite as clean):
* monitor the connection process
* send an exit signal with reason `normal` to the connection
* waiting for the down message
  * if the down message has `normal` for reason, `ok` is returned
  * if the down message has any other reason (eg `noproc`), an exit exception is raised with the reason from the down message
  * if no down message is received within the specified timeout, the connection is demonitored and an exit exception with reason `timeout` is raised (this does not mean that the connection process will not stop at all, only that it didn't within the given timeout)

The behavior of `mysql:stop/{1,3}` differs from what can be achieved with `exit/2` in that it is _synchronous_, ie the calling process will block until it has finished or timed out. To easily enable asynchronous stopping, a convenience function `mysql:stop_async/1` is in the PR also, which will call `stop` with timeout `infinity` in a spawned process and immediately return `ok`. Other than with synchronous `stop`, none of the errors will reach the process calling `stop_async`, and there is no easy way (apart from sth like `erlang:is_process_alive/1`) to determine if and when the connection closed.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -25,7 +25,8 @@
 %% gen_server is locally registered.
 -module(mysql).
 
--export([start_link/1, query/2, query/3, query/4, query/5,
+-export([start_link/1, stop/1, stop/2,
+         query/2, query/3, query/4, query/5,
          execute/3, execute/4, execute/5,
          prepare/2, prepare/3, unprepare/2,
          warning_count/1, affected_rows/1, autocommit/1, insert_id/1,
@@ -154,6 +155,51 @@ start_link(Options) ->
         _ -> ok
     end,
     Ret.
+ 
+%% @see stop/2.
+-spec stop(Conn) -> ok
+    when Conn :: connection().
+stop(Conn) ->
+    stop(Conn, infinity).
+
+%% @doc Stops a connection process and closes the connection. The
+%% process calling `stop' will be blocked until the connection
+%% process stops or the given timeout expires.
+%%
+%% If the connection is not stopped within the given timeout,
+%% an exit exception is raised with reason `timeout'.
+%%
+%% If the connection process exits with any other reason than `normal',
+%% an exit exception is raised with that reason.
+-spec stop(Conn, Timeout) -> ok
+    when Conn :: connection(),
+         Timeout :: timeout().
+stop(Conn, Timeout) ->
+    case erlang:function_exported(gen_server, stop, 3) of
+        true -> gen_server:stop(Conn, normal, Timeout);            %% OTP >= 18
+        false -> backported_gen_server_stop(Conn, normal, Timeout) %% OTP < 18
+    end.
+
+-spec backported_gen_server_stop(Conn, Reason, Timeout) -> ok
+    when Conn :: connection(),
+         Reason :: term(),
+         Timeout :: timeout().
+backported_gen_server_stop(Conn, Reason, Timeout) ->
+    Monitor=monitor(process, Conn),
+    exit(Conn, Reason),
+    receive
+        {'DOWN', Monitor, process, Conn, Reason} ->
+            ok;
+        {'DOWN', Monitor, process, Conn, UnexpectedReason} ->
+            exit(UnexpectedReason)
+    after Timeout ->
+        exit(Conn, kill),
+        receive
+            {'DOWN', Monitor, process, Conn, killed} ->
+                exit(timeout)
+        end
+    end.
+
 
 %% @see query/5.
 -spec query(Conn, Query) -> Result
--- a/test/ssl_tests.erl
+++ b/test/ssl_tests.erl
@@ -47,12 +47,4 @@ common_basic_check(ExtraOpts) ->
 
 common_conn_close() ->
     Pid = whereis(tardis),
-    process_flag(trap_exit, true),
-    exit(Pid, normal),
-    receive
-        {'EXIT', Pid, normal} -> ok
-    after
-        5000 -> error({cant_stop_connection, Pid})
-    end,
-    process_flag(trap_exit, false).
-
+    mysql:stop(Pid, 5000).
--- a/README.md
+++ b/README.md
@@ -79,8 +79,8 @@ end
 {ok, [<<"SLEEP(5)">>], [[1]]} =
     mysql:query(Pid, <<"SELECT SLEEP(5)">>, 1000),
 
-%% Close the connection; make it exit normally
-exit(Pid, normal).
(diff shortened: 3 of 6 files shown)


zuiderkwast:
Good idea!

I guess many people had issues with missing `gen_server:stop`. Do you have a reference to a discussion on erlang-questions or bugs.erlang.org or github? It would be interesting to know the background.

For async stop, do we really need that? Maybe calling exit can be enough for that? To compare: We don't have an async start function.

Regarding the code itself, I suppose you will move the mimicking code to a helper function and such cleanup...

juhlig:
> I guess many people had issues with missing `gen_server:stop`. Do you have a reference to a discussion on erlang-questions or bugs.erlang.org or github? It would be interesting to know the background.

Sorry, I don't, but I admit that I haven't looked far, either ;)

> For async stop, do we really need that? Maybe calling exit can be enough for that? To compare: We don't have an async start function.

No, we don't _need_ it ;) Just an idea/convenience thing ;) I'll remove it again ;)

Async start makes little sense as you do need the connection pid (with async stop OTOH you don't need anything from it), and there is not a good way to do that, at least not a general one.

> Regarding the code itself, I suppose you will move the mimicking code to a helper function and such cleanup...

Sure, this is just a proof of concept/WIP for your perousal as usual, to see if you like the idea. I'll make it pretty tomorrow ;) Tests and docs are missing, too.

zuiderkwast:
> this is just a proof of concept/WIP for your perousal as usual, to see if you like the idea

Sounds good! Yes, I like the idea! Except for async stop :-)

> Async start makes little sense as you do need the connection pid (with async stop OTOH you don't need anything from it), and there is not a good way to do that, at least not a general one.

Sure. For async start, I was rather thinking about how for example `mariaex` has async connect and also reconnect to DB if the connection is lost, without restarting the gen_server process. (I don't like this approach very much because the application may depend on state which is lost on reconnection to the DB.)

We can use sync stop in existing test cases if you want. Many of those exit calls can be replaced by stop. Also the example in the README.

juhlig:
Ok, so I removed `stop_async`, added docs, and moved the native (`gen_server:stop`) and emulation code into separate functions. Tests still pending, I'll see to that tomorrow.

src/mysql.erl
```diff
@@ -155,37 +155,47 @@ start_link(Options) ->
...
-                 receive
-                     {'DOWN', Monitor, process, Conn, normal} ->
-                         ok;
-                     {'DOWN', Monitor, process, Conn, Reason} ->
-                         exit(Reason)
-                 after Timeout ->
-                     demonitor(Monitor, [flush]),
-                     exit(timeout)
-                 end
+    stop(erlang:function_exported(gen_server, stop, 3), Conn, Timeout).
```
zuiderkwast: Maybe it's a matter of personal taste, but I was anticipating something like

```Erlang
    case erlang:function_exported(gen_server, stop, 3) of
        true  -> gen_server:stop(Conn, normal, Timeout);           %% OTP >= 18
        false -> backported_gen_server_stop(Conn, normal, Timeout) %% OTP < 18
    end
```

If you don't like my suggestion, I'm not going to be picky about it.

juhlig: Nah, either one is fine with me :) Did it.

This should be more or less the final version of it. In the tests, the `exit(Pid, normal)` calls have been swapped out for `mysql:stop`, which should provide a lot of coverage for it already. If you want, I can add some more tests dedicated to stopping to it, though.

I followed `gen_server:stop` through the OTP code, and found it is using `gen:stop`, which is using `proc_lib:stop`, which is using `sys:terminate`, all of which appeared with OTP 18, so that kinda explains why it didn't appear earlier. Following from there is harder as it starts sending messages around, and I was running out of time, so I stopped there ;) I guess the backported `gen_server;stop` is not a full backport, but as close as I could make it.

What `proc_lib:stop` does when it hits the timeout is kill the process. My former iterations of the backported function didn't do that, and I put that in with my last commit. What this means is that it will take down the process linked to the connection process, also, unless it is trapping exits. While I'm not sure that this is a very great thing, it is what `gen_server:stop` ultimately does, so the backport needs to do it as well in order to avoid inconsistent behavior between pre- and post-18 OTP releases.

juhlig: Oh, huh, the tests fail. Looks like I messed something up with the commit :P

juhlig: Ah, better :) Strange, how could that have ever worked locally :P

test/ssl_tests.erl
```diff
@@ -47,12 +47,4 @@ common_basic_check(ExtraOpts) ->
...
-    process_flag(trap_exit, true),
-    exit(Pid, normal),
-    receive
-        {'EXIT', Pid, normal} -> ok
-    after
-        5000 -> error({cant_stop_connection, Pid})
-    end,
-    process_flag(trap_exit, false).
-
+    mysql:stop(Pid, 5000).
```
zuiderkwast: It's great that this trap_exit crap can be removed!

src/mysql.erl
zuiderkwast: I can't see what you did here since you force-pushed, but I don't need to know...

zuiderkwast: > What `proc_lib:stop` does when it hits the timeout is kill the process. My former iterations of the backported function didn't do that, and I put that in with my last commit.

Good investigation! Kill is good. It's what a supervisor does when a child doesn't terminate within the specified time. A supervisor, poolboy and similar would trap_exit, so no worries for those.

zuiderkwast:
> In the tests, the `exit(Pid, normal)` calls have been swapped out for `mysql:stop`, which should provide a lot of coverage for it already. If you want, I can add some more tests dedicated to stopping to it, though.

This is great and enough. No need for any extra tests for stop.

Only the README examples could use stop now. I can't see that anything else needs to be done. It all looks great!

juhlig:
Ah, I forgot that, done. I'll remove the WIP from the title now, so it's good to go :)

PR #112, opened 2019-03-15
Fix the bug caused by delayed sub binary creation optimization in OTP…

GeraldXv:
Fix the bug caused by delayed sub binary creation optimization in OTP 21.0.


zuiderkwast:
Awesome! Thanks!

Issue #66, opened 2017-04-14
The whole application is shutdowned when using illegal parameter for query


johnzeng:
```
(eLockAccount@johnzeng-virtual-machine)1> mysql:query(pool_mysql, <<"select * from account where id =?">>, [undefine]).

=ERROR REPORT==== 14-Apr-2017::11:05:01 ===
** Generic server <0.1261.0> terminating
** Last message in was {param_query,<<"select * from account where id =?">>,
                                    [undefine]}
** When Server state == {state,[5,5,54],
                               95,#Port<0.3137>,"localhost",3306,"root",
                               "Removed",true,infinity,infinity,60000,0,
                               2,0,0,0,undefined,
                               {dict,0,16,16,8,80,48,
                                     {[],[],[],[],[],[],[],[],[],[],[],[],[],
                                      [],[],[]},
                                     {{[],[],[],[],[],[],[],[],[],[],[],[],[],
                                       [],[],[]}}},
                               empty,false}
** Reason for termination ==
** {function_clause,[{mysql_protocol,encode_param,
                                     [undefine],
                                     [{file,"src/mysql_protocol.erl"},
                                      {line,745}]},
                     {lists,map,2,[{file,"lists.erl"},{line,1238}]},
                     {mysql_protocol,execute,5,
                                     [{file,"src/mysql_protocol.erl"},
                                      {line,161}]},
                     {mysql,execute_stmt,4,
                            [{file,"src/mysql.erl"},{line,804}]},
                     {gen_server,try_handle_call,4,
                                 [{file,"gen_server.erl"},{line,629}]},
                     {gen_server,handle_msg,5,
                                 [{file,"gen_server.erl"},{line,661}]},
                     {proc_lib,init_p_do_apply,3,
                               [{file,"proc_lib.erl"},{line,240}]}]}
** exception exit: {{function_clause,[{mysql_protocol,encode_param,
                                                      [undefine],
                                                      [{file,"src/mysql_protocol.erl"},{line,745}]},
                                      {lists,map,2,[{file,"lists.erl"},{line,1238}]},
                                      {mysql_protocol,execute,5,
                                                      [{file,"src/mysql_protocol.erl"},{line,161}]},
                                      {mysql,execute_stmt,4,[{file,"src/mysql.erl"},{line,804}]},
                                      {gen_server,try_handle_call,4,
                                                  [{file,"gen_server.erl"},{line,629}]},
                                      {gen_server,handle_msg,5,
[...]

PR #113, opened 2019-03-28
Parameter validation for prepared statements and parameterized queries

juhlig:
The connection process crashed when invalid parameters were given to a parameterized query or to the execution of a prepared statement (see #66).

The additions in this PR will check the parameters for suitability and raise an `error` exception with reason `badarg` if one of them is invalid.

I'm not entirely happy with the unicode check, but I'm currently unable to come up with another safe way to do that.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -357,7 +357,13 @@ query(Conn, Query, Params, FilterMap) when (Params == no_params orelse
 query(Conn, Query, no_params, FilterMap, Timeout) ->
     query_call(Conn, {query, Query, FilterMap, Timeout});
 query(Conn, Query, Params, FilterMap, Timeout) ->
-    query_call(Conn, {param_query, Query, Params, FilterMap, Timeout}).
+    case mysql_protocol:valid_params(Params) of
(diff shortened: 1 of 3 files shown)


src/mysql.erl
```diff
@@ -731,3 +743,39 @@ query_call(Conn, CallReq) ->
...
+    true;
+is_valid_param(Value) when is_binary(Value) ->
+    true;
+is_valid_param(Value) when is_list(Value) ->
+    try
+        unicode:characters_to_binary(Value)
+    of
+        Value1 when is_binary(Value1) ->
+            true;
+        _ ->
```
zuiderkwast: I think this is the only way and thus the right way to do unicode validation.

The catch-all variable here could called `_ErrorOrIncomplete` instead of just `_`, for clarity.

juhlig: Ah, good idea :) I moved it `valid_params` into `mysql_protocol` and set up a test for it there. This way, it is closer to the related `encode_param` function, too.

I also renamed the catch-all variable as you suggested.

Issue #54, opened 2016-11-23
How to skip checking statements  in prepare config


newsdy:
hi, i use mysql-otp with a pool named pooler, url:https://github.com/seth/pooler

a question occured as that:
    the pooler call mysql:start_link() and get the pid of mysql connection, the example config is that:
```
{pooler, [
           {pools, [
                    [{name, logger_pool},
                     {max_count, 50},
                     {init_count, 20},
                     {start_mfa,
                      {mysql, start_link, [[{host,"127.0.0.1"}, {port, 3306}, {user, "mycomm"}, {password, "mycomm123"}, {database, "c8config"},
                                            {prepare,
                                             [
                                              {'heartbeat.select', "select 1 from dual"}
                                             ]
                                            }
                                           ]]}}]
                   ]}
          ]}
```

when the statements are wrong, such as unknown column, mysql-otp throw an error of badmatch in 

```
Prepare = proplists:get_value(prepare, Options, []),
            lists:foreach(fun ({Name, Stmt}) ->
                              {ok, Name} = mysql:prepare(Pid, Name, Stmt)
                          end,
                          Prepare);
```

`pool 'logger_pool' failed to start member:{error,{'EXIT',{{badmatch,{error,{1054,<<"42S22">>,<<"Unknown column 'work_start_time11' in 'where clause'">>}}},[{mysql,'-start_link/1-fun-1-',2,[{file,"/root/workspace/cti_dialing_svr_test/_build/default/lib/mysql/src/mysql.erl"},{line,163}]},{lists,foreach,2,[{file,"lists.erl"},{line,1336}]},{mysql,start_link,1,[{file,"/root/workspace/cti_dialing_svr_test/_build/default/lib/mysql/src/mysql.erl"},{line,162}]},{supervisor,do_start_child_i,3,[{file,"supervisor.erl"},{line,330}]},{supervisor,handle_call,3,[{file,"supervisor.erl"},{line,355}]},{gen_server,try_handle_call,4,[{file,"gen_server.erl"},{line,607}]},{gen_server,handle_msg,5,[{file,"gen_server.erl"},{line,639}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,237}]}]}}}`

Pooler thought the connection was created failed, but mysql connection was created success.
I have a timer to execute stmt, so pooler always created a new connection.After a while the mysql connections reached the upper limit.
Use `show processlist` in mysql
I think it's not Pooler's fault, right?
[...]

PR #115, opened 2019-03-29
Stop connection process when post-connect queries or prepares fail

juhlig:
This PR addresses the issue outlined in #54.

When one of the post-connect queries or prepares fails, the connection process will now be stopped again via `mysql:stop`, and an `exit` exception will be raised with the error code and message caused by the faulty query/prepare. The same will happen when there are failures in queries/prepares given to `change_user`.

It may be worth mentioning that all queries before a failing one will have been executed when the error occurs, ie that this is not atomic. This is nothing new with this PR, though. Destructive writes in on-connect/-change-user queries are generally a bad idea, I'd say.

I also corrected a slight inconsistency in the `mysql_change_user_tests` that have actually nothing to do with the issue at hand, but I didn't think it worth a separate PR.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -714,22 +714,31 @@ encode(Conn, Term) ->
 
 %% @doc Executes the given queries and prepares the given statements after a
 %% connection has been made.
+%%
+%% If any of the queries or prepares fails, the connection is closed and an
+%% exception is raised.
 -spec execute_after_connect(connection(), [iodata()], [{atom(), iodata()}])
     -> ok.
 execute_after_connect(Conn, Queries, Prepares) ->
-    lists:foreach(fun (Query) ->
-                      case query(Conn, Query) of
-                          ok -> ok;
-                          {ok, _} -> ok;
-                          {ok, _, _} -> ok
-                      end
-                  end,
-                  Queries),
-    lists:foreach(fun ({Name, Stmt}) ->
-                      {ok, Name} = prepare(Conn, Name, Stmt)
-                  end,
-                  Prepares),
-    ok.
+    try
+        lists:foreach(fun (Query) ->
+                          case query(Conn, Query) of
+                              ok -> ok;
+                              {ok, _} -> ok;
+                              {ok, _, _} -> ok
+                          end
+                      end,
+                      Queries),
+        lists:foreach(fun ({Name, Stmt}) ->
+                          {ok, Name} = prepare(Conn, Name, Stmt)
+                      end,
+                      Prepares),
+        ok
+    catch
+        ?EXCEPTION(Class, Reason, Stacktrace) ->
+            catch stop(Conn, ?default_connect_timeout),
+            erlang:raise(Class, Reason, ?GET_STACK(Stacktrace))
+    end.
 
 %% @doc Makes a gen_server call for a query (plain, parametrized or prepared),
 %% checks the reply and sometimes throws an exception when we need to jump out
--- a/test/mysql_change_user_tests.erl
+++ b/test/mysql_change_user_tests.erl
@@ -29,7 +29,7 @@
 %% Ensure that the current user can be changed to another user
 %% when given correct credentials.
 correct_credentials_test() ->
-    Pid = connect_db(?user1),
+    Pid = connect_db(?user1, ?password1),
     ?assertEqual(ok, mysql:change_user(Pid, ?user2, ?password2)),
     ?assert(is_current_user(Pid, ?user2)),
     mysql:stop(Pid),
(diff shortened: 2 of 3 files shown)


src/mysql.erl
```diff
@@ -712,19 +712,30 @@ encode(Conn, Term) ->
...
 execute_after_connect(Conn, Queries, Prepares) ->
     lists:foreach(fun (Query) ->
                       case query(Conn, Query) of
                           ok -> ok;
                           {ok, _} -> ok;
-                          {ok, _, _} -> ok
+                          {ok, _, _} -> ok;
+                          {error, Reason} ->
+                              stop(Conn),
+                              exit(Reason)
```
zuiderkwast: What's the reasoning for raising `exit` vs `error`?

What if the query raises an exception (e.g. tcp error or badarg such as if Query isn't iodata), is that worth catching?

juhlig: > What's the reasoning for raising `exit` vs `error`?

Hm, I usually do that "by feeling". I have been searching if there was a rule for when to use which, but even a post by Robert Virding didn't shed much light on it, except that `exit` should be used to hint to the calling process that it may want to exit itself. So my thinking was, when a process wants a database connection, it usually _needs_ it, ie can't continue without it.

I can of course change it if you would prefer `error` here.

> What if the query raises an exception (e.g. tcp error or badarg such as if Query isn't iodata), is that worth catching?

Good point, I only thought of a query/prepare being invalid in terms of "invalid for the database", not for `mysql-otp`. I'll add a try around the whole thing then.

zuiderkwast:
I messed with the commit history last night. Sorry for that! (It started with a spelling correction in a commit message, then some line breaking of other commit messages. I shoudn't have done that, I realized after.)

src/mysql.erl
zuiderkwast: I think `exit` is suitable for an error you can't recover from by catching it or something like that. If a query crashes an existing connection process, it's an obvious reason for exit, but in start_link, I don't know. Isn't it possible to recover by calling start_link again? From the caller's point of view, it's looks like we just failed to create the connection process.

If we go for exit, it should probably be an exit the try-catch case as well, regardless of what kind of exception was raised because the result is that we stop the connection here too.

Regarding the the Reason, to me it's not obvious that reusing error reasons as exit reasons is the right thing to do. Is it? Another possibility is to raise a general message such as failed to start server and possibly also log the SQL error using logger (or error_logger for older OTPs).

A third possibility is to re-raise the same exception again using the same stacktrace (using erlang:raise/3) after stopping the connection in the catch clause and let errors in query or prepare just raise a case_clause error (i.e. just not adding the `{error, Reason}` case clauses for these).

I'm not sure which one I prefer. Maybe the third one. What do you think?

juhlig:
No problem, got it sorted I think :)

src/mysql.erl
juhlig: Sounds reasonable to me :)

juhlig: Ok, have a look please :)

src/mysql.erl
```diff
@@ -712,22 +712,31 @@ encode(Conn, Term) ->
...
+                      Queries),
+        lists:foreach(fun ({Name, Stmt}) ->
+                          {ok, Name} = prepare(Conn, Name, Stmt)
+                      end,
+                      Prepares),
+        ok
+    catch
+        ?EXCEPTION(Class, Reason, Stacktrace) ->
+            catch stop(Conn),
+            erlang:raise(Class, Reason, ?GET_STACK(Stacktrace))
```
zuiderkwast: Yes, exactly, this is what I was thinking. So you agree this is a good way?

Just one thing: Perhaps it would be good to use stop/2 with a timeout (any value less than infinity, perhaps just reusing `?default_connect_timeout`) to make sure this can't hang the start_link call forever. What do you think?

juhlig: Yes, I agree.  :)

A timeout is good, I'd say.

But, if it expires, it will also kill the calling process, since they are linked at this point already. But I guess expiry is very unlikely, and I'm not sure if this could be a problem even if it happens.

We could also use a short-lived process to do the `mysql:stop` call and unlink the connection process from the calling process.

zuiderkwast: I too think expiry is very unlikely. Not worth adding the complexity of a short-lived process IMO. Adding a timeout is very simple though. It kills the calling process (unless it's trapping exits) but that's better than hanging.

juhlig: Ok, done :)

juhlig:
I have a feeling that I will move that into the connection process one day, though. It has kind of a bolted-on feeling the way it is now ;)

Issue #114, opened 2019-03-28
It looks like you support Unix domain sockets


michaellenaghan:
Not a bug, a feature! 

Please forgive the Elixir dialect, but in 1.3.1, which is what's [currently on Hex](https://hex.pm/packages/mysql), it's possible to use a Unix domain socket by doing this:

```
iex(2)> {:ok, pid} = :mysql.start_link([
...(2)>   host: {:local, "/opt/local/var/run/mysql56/mysqld.sock" |> to_charlist},
...(2)>   port: 0,
...(2)>   database: "some_db" |> to_charlist, 
...(2)>   user: "some_user" |> to_charlist, 
...(2)>   password: "some_password" |> to_charlist
...(2)> ])
{:ok, #PID<0.1089.0>}
```

Worth mentioning somewhere? (Apologies if this is obvious to an old Erlang hand; it wasn't to this very new Elixir one!)

PS Possible to update Hex?

michaellenaghan:
To keep this working I think the [unreleased change to the `host` spec](https://github.com/mysql-otp/mysql-otp/blob/master/src/mysql.erl#L126) will have to change again:

```
    {host, inet:socket_address() | inet:hostname()} | {port, integer()} |
```

Maybe something like this?

```
    {host, inet:socket_address() | inet:hostname() | {local, inet:local_address()}} | {port, integer()} |
```

 (Untested.)

zuiderkwast:
Oh, that's great news! I didn't know! :-) I think unix domain socket was added to OTP in one of the recent versions, so it might just have started working out of nowhere.

We should have at least one test case connecting to local host using the unix domain socket. And fix the spec as mentioned.

Btw, I don't think you need to do `to_charlist`. For the local socket I'm not sure, but for user, password and database, a binary string should work fine.

juhlig:
> Oh, that's great news! I didn't know! :-) I think unix domain socket was added to OTP in one of the recent versions, so it might just have started working out of nowhere.

Yup, it was introduced in experimental state in OTP 19.0, if I'm not mistaken. Don't know when it went to official state.

> We should have at least one test case connecting to local host using the unix domain socket. And fix the spec as mentioned.

[...]

PR #116, opened 2019-03-30
Default port 0 for unix socket

juhlig:
In reference to #114, this changes the default port behavior if the host is a unix (`{local, ...}`) socket to use `0` instead of `3306`.

Is this style to your liking? It would be possible to override even a user-specified port in case of a `{local, ...}` host setting, but I don't think we should. Just provide a sensible default.

The test I will do later, I'm not on my usual dev machine now ;) I was thinking of checking OTP version first to see if it is >=19 as you suggested, then connecting via the usual `localhost`, query the database for `@@socket` (as this may be different on different OSes, and I can't think of another easy way to find that out), then use that to attempt a `{local, ...}` connect. Ok with you?

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -76,9 +76,11 @@
 %%       name. For details see the documentation for the first argument of
 %%       gen_server:start_link/4.</dd>
 %%   <dt>`{host, Host}'</dt>
-%%   <dd>Hostname of the MySQL database; default `"localhost"'.</dd>
+%%   <dd>Hostname of the MySQL database. Since OTP version 19, it is also
+%%       possible to specify a local (Unix) Socket by specifying
+%%       `{local, SocketFile}'. Default `"localhost"'.</dd>
 %%   <dt>`{port, Port}'</dt>
-%%   <dd>Port; default 3306 if omitted.</dd>
+%%   <dd>Port; default 3306 for non-local or 0 for local (Unix) sockets.</dd>
 %%   <dt>`{user, User}'</dt>
 %%   <dd>Username.</dd>
 %%   <dt>`{password, Password}'</dt>
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -61,7 +61,13 @@
 init(Opts) ->
     %% Connect
     Host           = proplists:get_value(host, Opts, ?default_host),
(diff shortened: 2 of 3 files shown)


zuiderkwast:
Yes, looks good.

You mean the Unix socket is not available on e.g. Windows? I was mainly concerned about Travis so I didn't think about this, but you're right, it's better to check `@@socket` as you suggest.

juhlig:
> You mean the Unix socket is not available on e.g. Windows? I was mainly concerned about Travis so I didn't think about this, but you're right, it's better to check `@@socket` as you suggest.

I'm not really certain, as I only have Windows on my working machine but no MySQL on it, and I can't just install stuff without a work-related reason there ;) If it _is_ set, it should be something like `null`, or empty there.

My main reason for fetching it is that the socket file could be in different locations on OSes that _do_ support it. Eg, on Ubuntu it is `/var/run/mysqld/mysqld.sock`, on FreeBSD it is `/tmp/mysqld.sock` I think.

juhlig:
Ah, from the [MySQL docs on `socket`](https://dev.mysql.com/doc/refman/5.6/en/server-system-variables.html#sysvar_socket): "On Windows, this variable is the name of the named pipe that is used for local client connections. The default value is MySQL (not case-sensitive)."

juhlig:
Ok, I set up a test. It skips on OTP<19, and if the connect attempt returns eafnosupport (ie, windows).

zuiderkwast:
You're right, we need the socket path. No need to test this for Windows IMO. Do as you first planned. (And sorry for messing up the git history!)

juhlig:
Pushed it already, have a look ;)

The test checks for OTP>=19 first and...
* if `true`, fetches the socket path and tries to connect through it via `{local, ...}`
  * if the reply is `{ok, Pid}`, it executes a simple query for good measure, and the test is a success
  * if the reply is `{error, eafnosupport}` instead, the OS doesn't support the address family, and the test skips (this is probably Windows then, as it is the only mainstream OS that I can think of that doesn't support unix sockets, but who knows)
  * if the reply is something else (eg `{error, enoent}` -> socket file is missing), there is something messed up, so the test fails
* if `false`, skips

juhlig:
#116 should be closed by this.

PR #117, opened 2019-04-02
Ignore SERVER_STATUS_AUTOCOMMIT on handshake

imrivera:
In some cases, the server may return that AUTOCOMMIT is not enabled, but after login it can be enabled. Before this fix, that makes the connection crash.

### DIFF
--- a/src/mysql_protocol.erl
+++ b/src/mysql_protocol.erl
@@ -83,8 +83,11 @@ handshake_finish_or_switch_auth(Handshake = #handshake{status = Status}, Passwor
     {ok, ConfirmPacket, SeqNum1} = recv_packet(SockModule, Socket, SeqNum0),
     case parse_handshake_confirm(ConfirmPacket) of
         #ok{status = OkStatus} ->
-            %% check status, ignoring bit 16#4000, SERVER_SESSION_STATE_CHANGED.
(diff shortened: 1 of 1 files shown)


zuiderkwast:
Interesting! Under which conditions have you seen this happening? Which MySQL or MariaDB version?

imrivera:
Hi! I've seen this connecting to ProxySQL 1.4.14 that connects to Percona XtraDB cluster (galera) 5.6.41. In the server greeting we are getting a AUTOCOMMIT = TRUE, but after sending the credentials, in the response we are getting AUTOCOMMIT = FALSE.

It seems a "bug" in ProxySQL, as it's creating the response with a line like this:
`client_myds->myprot.generate_pkt_OK(true,NULL,NULL, _pid, 0,0,0,0,NULL);`
(StatusFlags is the 7th parameter).

Probably we'll try to file a pull request to ProxySQL also to fix this behaviour, but anyway. Other libraries/clients doesn't have any problem connecting to it. So maybe we should ignore that bit in the handshake.

zuiderkwast:
Sure, I'll merge it! I was just interested. Thanks for the background information!

imrivera:
Thank you :)

zuiderkwast:
Thanks for reporting and fixing it!

PR #123, opened 2019-04-24
Change the character set from utf8mb3 to utf8mb4 for servers that (should) support it.

### LINKED ISSUE #121: Consider changing the default character set to utf8mb4
Please consider changing the default character set [here](https://github.com/mysql-otp/mysql-otp/blob/f43adad8e0a50c8f8120c407ec820d35add2f219/include/protocol.hrl#L25):

```
-define(UTF8, 16#21). %% utf8_general_ci
```

to:

```
-define(UTF8, 16#2d). %% utf8mb4_general_ci
```

Otherwise characters off the basic multilingual plane (e.g., emoticons and many Asian characters) will, depending on the SQL mode, either silently be discarded or error:

```
>  :mysql.query(conn, "CREATE TABLE t (i INT UNSIGNED AUTO_INCREMENT PRIMARY KEY, c CHAR(20) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci);")
:ok

>  :mysql.query(conn, "select @@character_set_results, @@character_set_client, @@character_set_connection;")
{:ok, 
  ["@@character_set_results", "@@character_set_client", "@@character_set_connection"], 
  [["utf8", "utf8", "utf8"]]}

> :mysql.query(conn, "set session sql_mode = '';")  
:ok

> :mysql.query(conn, "insert into t (c) values (?);", ["!!!😃!!!"])           
11:41:25.669 [warn]  Warning 1366: Incorrect string value: '\xF0\x9F\x98\x83!!...' for column 'c' at row 1
 in insert into t (c) values (?);

:ok

> :mysql.query(conn, "select * from t where i = ?;", [:mysql.insert_id(conn)])
{:ok, ["i", "c"], [[1, "!!!????!!!"]]}

> :mysql.query(conn, "set session sql_mode = 'TRADITIONAL';")            
:ok  

> :mysql.query(conn, "insert into t (c) values (?);", ["!!!😃!!!"])      
{:error,
 {1366, "HY000",
  "Incorrect string value: '\\xF0\\x9F\\x98\\x83??...' for column 'c' at row 1"}}
```

Here's what that sequence looks like after making the change:

```
>  :mysql.query(conn, "select @@character_set_results, @@character_set_client, @@character_set_connection;")
{:ok, 
  ["@@character_set_results", "@@character_set_client", "@@character_set_connection"], 
  [["utf8mb4", "utf8mb4", "utf8mb4"]]}

> :mysql.query(conn, "insert into t (c) values (?);", ["!!!😃!!!"])           
:ok

> :mysql.query(conn, "select * from t where i = ?;", [:mysql.insert_id(conn)])
{:ok, ["i", "c"], [[1, "!!!😃!!!"]]}
```

(The SQL mode no longer matters since there's no longer a problem.)

In the meantime a workaround (or an alternative approach if you don't think `utf8mb4` should be the default) is to add this to the list of `connect` queries:

    "SET NAMES utf8mb4;"

michaellenaghan:
(Fixes #121)

### DIFF
--- a/src/mysql_protocol.erl
+++ b/src/mysql_protocol.erl
@@ -27,7 +27,7 @@
 %% @private
 -module(mysql_protocol).
 
--export([handshake/7, change_user/6, quit/2, ping/2,
+-export([handshake/7, change_user/7, quit/2, ping/2,
          query/4, query/5, fetch_query_response/3,
          fetch_query_response/4, prepare/3, unprepare/3,
          execute/5, execute/6, fetch_execute_response/3,
@@ -233,8 +233,9 @@ fetch_execute_response(SockModule, Socket, FilterMap, Timeout) ->
 
 %% @doc Changes the user of the connection.
 -spec change_user(module(), term(), iodata(), iodata(), binary(),
-                  undefined | iodata()) -> #ok{} | #error{}.
-change_user(SockModule, Socket, Username, Password, Salt, Database) ->
+                  undefined | iodata(), [integer()]) -> #ok{} | #error{}.
+change_user(SockModule, Socket, Username, Password, Salt, Database, 
+            ServerVersion) ->
     DbBin = case Database of
         undefined -> <<>>;
         _ -> iolist_to_binary(Database)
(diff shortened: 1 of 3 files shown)


src/mysql_protocol.erl
```diff
@@ -431,6 +432,14 @@ add_client_capabilities(Caps) ->
     ?CLIENT_MULTI_RESULTS bor
     ?CLIENT_PS_MULTI_RESULTS.
 
+-spec character_set([integer()]) -> integer().
+character_set(ServerVersion) when ServerVersion >= [5, 5, 3] ->
+    %% https://dev.mysql.com/doc/relnotes/mysql/5.5/en/news-5-5-3.html
```
zuiderkwast: Very good reference comment!

zuiderkwast:
Awesome! Thanks!

PR #125, opened 2019-04-24
Make buffer at least as large as recbuf if it wasn’t specifically defined.

### LINKED ISSUE #119: Possible performance tweaks?
Here's what the erlang docs [say](http://erlang.org/doc/man/inet.html#setopts-2) about the `buffer` socket option:

> For TCP it is recommended to have val(buffer) >= val(recbuf) to avoid performance issues because of unnecessary copying. ... val(buffer) is automatically set to the above maximum when recbuf is set. However, as the size set for recbuf usually become larger, you are encouraged to use getopts/2 to analyze the behavior of your operating system.
> 
> Note that this is also the maximum amount of data that can be received from a single recv call. If you are using higher than normal MTU consider setting buffer higher.

On my machine here's what I get by default:

```
iex(8)> :inet.getopts(elem(:sys.get_state(conn), 3), [:buffer])
{:ok, [buffer: 1460]}                                                           
iex(9)> :inet.getopts(elem(:sys.get_state(conn), 3), [:recbuf])     
{:ok, [recbuf: 8192]}
iex(10)> :inet.getopts(elem(:sys.get_state(conn), 3), [:sndbuf])            
{:ok, [sndbuf: 8192]}
```

If I'm reading the docs right the default `buffer` size should be larger?

The reason I'm raising this as an issue is that, as far as I can tell, and as you can see above, I have to break encapsulation in order to do this myself.

Thoughts?

Another possible optimization is to set `nodelay` to `true`; the default is `false`:

```
iex(11)> :inet.getopts(elem(:sys.get_state(conn), 3), [:nodelay])
{:ok, [nodelay: false]}
```

Here's the description:

> If `Boolean == true`, option `TCP_NODELAY` is turned on for the socket, which means that also small amounts of data are sent immediately.

Since `mysql-otp` only sends complete messages I'm not sure delaying makes sense.

michaellenaghan:
(Fixes #119)

This roughly follows the `myxql` approach: if the user defined a buffer size, use it, otherwise set it. Another option would be to set the buffer size unconditionally; if the user made the buffer small it would then be adjusted to the suggested size, and if they made it large it would remain unchanged.

### DIFF
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -94,8 +94,8 @@ init(Opts) ->
 
     %% If buffer wasn't specifically defined make it at least as
     %% large as recbuf, as suggested by the inet:setopts() docs.
-    BufferIsNotDefined = not proplists:is_defined(buffer, TcpOpts),
-    if BufferIsNotDefined ->
+    BufferIsDefined = proplists:is_defined(buffer, TcpOpts),
+    if not BufferIsDefined ->
         {ok, [{buffer, Buffer}, {recbuf, Recbuf}]} = inet:getopts(Socket0,
                                                                   [buffer, 
                                                                    recbuf]),
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -96,9 +96,8 @@ init(Opts) ->
     %% large as recbuf, as suggested by the inet:setopts() docs.
     BufferIsDefined = proplists:is_defined(buffer, TcpOpts),
     if not BufferIsDefined ->
-        {ok, [{buffer, Buffer}, {recbuf, Recbuf}]} = inet:getopts(Socket0,
-                                                                  [buffer, 
-                                                                   recbuf]),
+        {ok, [{buffer, Buffer}]} = inet:getopts(Socket0, [buffer]),
+        {ok, [{recbuf, Recbuf}]} = inet:getopts(Socket0, [recbuf]),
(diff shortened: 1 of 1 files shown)


michaellenaghan:
> Minor: The negated variable can be avoided by moving the not to the if...

You're right; good suggestion.

> I guess it is safe to assume the options are always returned in this order.

You're right; good catch.

I tracked down [this](https://github.com/erlang/otp/blob/d6285b0a347b9489ce939511ee9a979acd868f71/lib/compiler/test/compilation_SUITE_data/convopts.erl#L41) test in Erlang that makes the same assumption. I tracked down the underlying implementation [here](https://github.com/erlang/otp/blob/50a2582b39507149452d298129616a7d5feafbf2/erts/emulator/drivers/common/inet_drv.c#L7562) in C; it basically maps over the list. Neither is definitive, just indicative.

It would be trivially easy to request the opts in two lines rather than one. I may do that even though it may be unnecessary.

zuiderkwast:
> I tracked down this test in Erlang that makes the same assumption. I tracked down the underlying implementation here in C; it basically maps over the list. Neither is definitive, just indicative.

Wow, lots of work. I assumed it would work and I didn't suggest to change it. :-)

> It would be trivially easy to request the opts in two lines rather than one. I may do that even though it may be unnecessary.

Ok, sure. Calling `proplists:get_value/2` on the result would be another possibility.

Good job!

PR #126, opened 2019-04-24
add `tcp_options` to `start_link/1` spec

juhlig:
In response to https://github.com/mysql-otp/mysql-otp/issues/119#issuecomment-484740620: you're right, fixed it :)

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -135,7 +135,8 @@
                    {queries, [iodata()]} |
                    {query_timeout, timeout()} |
                    {found_rows, boolean()} |
-                   {query_cache_time, non_neg_integer()},
+                   {query_cache_time, non_neg_integer()} |
+                   {tcp_options, [gen_tcp:connect_option()]},
          ServerName :: {local, Name :: atom()} |
                        {global, GlobalName :: term()} |
                        {via, Module :: atom(), ViaName :: term()},


zuiderkwast:
Good!

I think `{ssl, SslOptions}` is missing too, both in the spec and in the doc above.

juhlig:
> I think `{ssl, SslOptions}` is missing too, both in the spec and in the doc above.

Yes, but this one is tricky in the spec... the type is changing, in 21.3 it is `ssl:tls_client_option()`, before it was `ssl:connect_option()` or something like that. What do you suggest?

zuiderkwast:
Conditionally defined type? E.g.
```Erlang
-if OTP_RELEASE >= 21.3.
-define(SLL_OPTION_TYPE, ssl:connect_option()).
-else.
-define(SLL_OPTION_TYPE, ssl:tls_client_option()).
-endif.
```

... or using `-type` instead of `-define`.

juhlig:
> -if OTP_RELEASE >= 21.3.

This won't work, since OTP_RELEASE contains only the major version, `21`.

It seems the `ssl` module is currently changing a lot, so this may not even be the end of it. I know of another type that has moved around the `ssl` modules a lot, see here https://github.com/ninenines/ranch/issues/218 and here https://github.com/ninenines/ranch/pull/220. So maybe we should follow suit and just support dialyzer in the most recent version?

zuiderkwast:
Or we could type it as the general type `{ssl, any()}`.

juhlig:
Ok, see #129.

PR #127, opened 2019-04-24
Enable nodelay by default.

### LINKED ISSUE #119: Possible performance tweaks?

michaellenaghan:
(Fixes #119)

### DIFF
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -89,7 +89,8 @@ init(Opts) ->
     end,
 
     %% Connect socket
-    SockOpts = [binary, {packet, raw}, {active, false} | TcpOpts],
+    SockOpts = [binary, {packet, raw}, {active, false}, {nodelay, true}
+                | TcpOpts],
     {ok, Socket0} = SockMod0:connect(Host, Port, SockOpts),
 
     %% If buffer wasn't specifically defined make it at least as


zuiderkwast:
Nice! I removed "by default" in the commit message, because it's not only by default but always. There's no need to be able to turn it off, right?

michaellenaghan:
I saw code somewhere that suggested `gen_tcp` takes the last option in the list rather than the first. If that's true then users could, in fact, override the setting. 

To answer your question more directly, though: no, I can't think of a reason to do that given the way MySQL's protocol works.

PR #128, opened 2019-04-26
WIP: separate user from connection API

juhlig:
The user API is intertwined with the connection API, so it is often impossible to make changes in one without having to change the other.

This PR removes all calls to `gen_server` from the `mysql` module and moves them into functions in the `mysql_conn` module, along with some of the surrounding logic.

So, the `mysql` module doesn't need to know that `mysql_conn` is a `gen_server`, doesn't need to know the structure of the messages it expects, gets rid of much of the logic, and thus becomes more of a real front-end API module. Most of what it now does is just pass through calls to `mysql_conn`. The large edoc comments that don't concern the actual code and are more of a user guide automatically fit better.


juhlig:
I don't think new tests are needed, btw, as the current ones automatically cover this.

You might notice that there are no specs on the `mysql_conn` API functions. I'm not sure how best to write them, should I use the types from `mysql`, like `mysql:connection()`, or move the types into `mysql_conn` and use them in `mysql` like `mysql_conn:connection()`? From a documentation perspective, I tend to the former; OTOH, considering the aspect that `mysql` is now just the "pretty face" on the "real thing", I tend to the latter... What do you think?

juhlig:
> The on-connect queries can be moved inside mysql_conn regardless of this change, right?

Yes they can, the two are completely separate things. I just would like to have this one out of the way one way or other to avoid unnecessary conflicts in the future.

For the rest of what you said, I agree in this specific case (the "case" being `mysql-otp`), but in a more general perspective I would vehemently disagree ;) Since this project is the only thing we work on together, we don't need to sort this out at all, and I'll just follow your preference. I'll try to explain my view of things, though.

> Duplicating the functions of mysql in mysql_conn. I normally don't like this sort of internal interface functions. It's adding boilerplate and longer call chains are harder to follow.

I would argue that it is much easier to follow modules and functions around than messages between processes. Most IDEs can do that for you, and autocomplete their names, and tell you how many args they expect, maybe even show you a snippet of the documentation to see what those args should be etc; all because functions (the ones defined in modules, not the anonymous ones of course) are "static things". It can't do any of that with messages, as they and the processes they are being sent from and to are "runtime things".

> `mysql` was itself the gen_server callback module until some time ago, but I separated the two.

I know, I started using it just before that was done ;)

> They are both part of the same application, so I see no benefit in hiding the fact that mysql_conn is a gen_server callback module from the mysql module. On the contrary, I like the fact that mysql_conn is _only_ a callback module.

The intention is not about _hiding_ it, though that is a (IMO nice) side effect, but so you can act more like "Do <thing> for me. I don't care what you really are and how you do it, just _do_ it", you know what I mean? Moreover, with all or at least most of the related parts in one place, there is only one place you have to make changes in, and if something doesn't work after you have made those, you know where to look first.

Anyway, I will close this PR then, and will start work on moving the on-connect stuff next week. Have a nice weekend :)

PR #129, opened 2019-05-02
add ssl options to start_link/1 spec and doc

juhlig:
This PR adds `ssl` options to the `start_link/1` spec and docs, as discussed in #126. I used `term()` instead of the proposed `any()` because the has an "it doesn't really matter" air to it ;)

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -121,6 +121,8 @@
 %%   <dd>Additional options for `gen_tcp:connect/3'. You may want to set
 %%       `{recbuf, Size}' and `{sndbuf, Size}' if you send or receive more than
 %%       the default (typically 8K) per query.</dd>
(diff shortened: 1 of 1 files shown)


src/mysql.erl
```diff
@@ -136,7 +138,8 @@
                    {query_timeout, timeout()} |
                    {found_rows, boolean()} |
                    {query_cache_time, non_neg_integer()} |
-                   {tcp_options, [gen_tcp:connect_option()]},
+                   {tcp_options, [gen_tcp:connect_option()] |
+                   {ssl, term()}},
```
zuiderkwast: Curly brace nesting error. The ssl options should be the value of the tcp_options.

Apart from that, good!

juhlig: oh, ok, fixed it =^^=

PR #134, opened 2019-09-23
Treat empty params list as no_params

suexcxine:
I passed in empty list params like this:
```
query(Db, Sql) ->                                                                                                                                                                         
    query(Db, Sql, []).                                                                                                                                                                   
                                                                                                                                                                                          
query(Db, Sql, Params) ->                                                                                                                                                                 
    case mysql_poolboy:query(Db, Sql, Params) of                                                                                                                                          
        {error, Reason} ->                                                                                                                                                                
            lager:error("mysql error: ~p, sql: ~s", [Reason, Sql]),                                                                                                                       
            {error, Reason};                                                                                                                                                              
        Result ->                                                                                                                                                                         
            Result                                                                                                                                                                        
    end.
```
and resulted in 
"Can't create more than max_prepared_stmt_count statements"

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -352,7 +352,8 @@ query(Conn, Query, Params, FilterMap) when (Params == no_params orelse
          Params :: no_params | [term()],
          FilterMap :: no_filtermap_fun | query_filtermap_fun(),
          Result :: query_result().
-query(Conn, Query, no_params, FilterMap, Timeout) ->
+query(Conn, Query, Params, FilterMap, Timeout) when Params =:= no_params; 
+                                                    Params =:= [] ->
     query_call(Conn, {query, Query, FilterMap, Timeout});
 query(Conn, Query, Params, FilterMap, Timeout) ->
     case mysql_protocol:valid_params(Params) of


zuiderkwast:
Interesting suggestion. It's not that simple to just accept this change though.

`query(Db, Dql, [])` is a prepared statement. It is actually the only way to create an unnamed prepared statement. This change makes it a plain query and not a prepared statement anymore.

If you have just a few fixed SQL queries repeated over and over in your program, prepared statements are supposed to be faster, because the SQL code and the values don't need to be parsed as plain text every time. On the other hand, prepared statements require one more roundtrip to the server. (The client first needs to send the command "prepare" to the server; then the server responds; then the client sends "execute" with or without params; then the server sends the results.)

If you're generating queries, you should rather use `query(Db, Sql)`. I suspect you are generating many different queries, since you get the error "Can't create more than max_prepared_stmt_count statements". If you need parameters, you can embed them in the SQL code using `mysql:encode/1` to avoid the extra round trip of prepared statements, e.g.

```Erlang
    Sql = "SELECT a, b, c FROM t WHERE x = " ++ mysql:encode(X),
    Result = mysql:query(Db, Sql).
```

Generating SQL code is not generally recommended though, so I'd like to keep the behaviour of query(Db, Sql, []), at least by default.

suexcxine:
Thank you for your reply. I understood the behavior now.
I'll just use query(Db, Sql) in my case.

PR #138, opened 2019-10-10
support 'COM_RESET_CONNECTION'

qingchuwudi:
A more lightweightt version of COM_CHANGE_USER that does about the same to clean up the session state, but:

1. it does not re-authenticate (and do the extra client/server exchange for that)
2. it does not close the connection

more : https://dev.mysql.com/doc/dev/mysql-server/8.0.11/page_protocol_com_reset_connection.html


qingchuwudi:
```bash
... ...
[0.086 s] ok
    mysql_tests: reset_connection_test...[0.020 s] ok
... ...
```
test pass

qingchuwudi:
It seems that **deadlock_test_** test failed.

```bash
=CRASH REPORT==== 10-Oct-2019::03:00:09.786565 ===
  crasher:
    initial call: mysql_conn:init/1
    pid: <0.213.0>
    registered_name: []
    exception exit: {application_process_died,<0.214.0>}
      in function  gen_server:handle_common_reply/8 (gen_server.erl, line 751)
    ancestors: [<0.80.0>]
    message_queue_len: 0
    messages: []
    links: []
    dictionary: []
    trap_exit: true
    status: running
    heap_size: 2586
    stack_size: 27
    reductions: 13247
  neighbours:
[0.031 s] ok
  transaction_tests: deadlock_test_ (Plain queries)...*skipped*
  =ERROR REPORT==== 10-Oct-2019::03:00:09.826753 ===
Error in process <0.226.0> with exit value:
{{implicit_commit,"UPDATE foo SET v = 2 WHERE k = 2"},
 [{mysql,query_call,2,[{file,"src/mysql.erl"},{line,726}]},
  {transaction_tests,'-deadlock_plain_queries/1-fun-1-',2,
                     [{file,"test/transaction_tests.erl"},{line,232}]},
  {mysql,execute_transaction,4,[{file,"src/mysql.erl"},{line,584}]},
  {transaction_tests,'-deadlock_plain_queries/1-fun-2-',2,
                     [{file,"test/transaction_tests.erl"},{line,230}]}]}
undefined
  *unexpected termination of test process*
::{{implicit_commit,"UPDATE foo SET v = 2 WHERE k = 2"},
   [{mysql,query_call,2,[{file,"src/mysql.erl"},{line,726}]},
    {transaction_tests,'-deadlock_plain_queries/1-fun-1-',2,
                       [{file,"test/transaction_tests.erl"},{line,232}]},
    {mysql,execute_transaction,4,[{file,"src/mysql.erl"},{line,584}]},
    {transaction_tests,'-deadlock_plain_queries/1-fun-2-',2,
                       [{file,"test/transaction_tests.erl"},{line,230}]}]}
```

src/mysql_conn.erl
```diff
@@ -314,9 +314,19 @@ handle_call({change_user, Username, Password, Options}, From,
...
-    Ok = mysql_protocol:reset_connnection(SockMod, Socket),
+    Result = mysql_protocol:reset_connnection(SockMod, Socket),
     setopts(SockMod, Socket, [{active, once}]),
-    {reply, ok, update_state(Ok, State)};
+    State1 = update_state(Result, State),
+    Reply = case Result of
+        #ok{} -> ok;
+        #error{} = E ->
+            %% 'COM_RESET_CONNECTION' is added in MySQL 5.7,
+            %% and return "Unkown command" in 5.6 (earlier version included).
```
zuiderkwast: Good error handling and comment!

src/mysql_conn.erl
```diff
@@ -314,9 +314,19 @@ handle_call({change_user, Username, Password, Options}, From,
...
+            State1#state.log_warnings andalso log_warnings(State1, "RESET CONNECTION"),
```
zuiderkwast: If there is an error, there are no warnings, but for an OK message there can be warnings from the server.

In other places, this logging is just after update_state, e.g. for CHANGE USER:

```Erlang
    State1 = update_state(Result, State),
    State1#state.warning_count > 0 andalso State1#state.log_warnings
        andalso log_warnings(State1, "CHANGE USER"),
```

If you look at `update_state/1`, where `warning_count` is updated from the result. If the packet is an error packet, the warning count is set to zero because the error packet doesn't contain any warning count.

test/mysql_tests.erl
```diff
@@ -182,8 +183,12 @@ reset_connection_test() ->
...
     ok = mysql:query(Pid, <<"INSERT INTO t (id, tx) VALUES (1, 'text 1')">>),
     ?assertEqual(1, mysql:insert_id(Pid)),  %% auto_increment starts from 1
-    ok = mysql:reset_connection(Pid),
-    ?assertEqual(0, mysql:insert_id(Pid)), %% insertid reset to 0
+    case mysql:reset_connection(Pid) of
+      ok ->
+        ?assertEqual(0, mysql:insert_id(Pid)); %% insertid reset to 0;
+      _Error ->
+        ?assertEqual(1, mysql:insert_id(Pid)) %% reset failed
+    end,
```
zuiderkwast: If reset_connection returns an error, how do we know it is because the server doesn't support it or because the function doesn't work as expected?

For >= 5.7.0, it should return `ok` and for < 5.7.0 maybe we don't care, or we accept both error and OK, but at we can check that it doesn't crash. To check the server version, something like `Version = db_version_string(Pid), case Version < [5, 7, 0] orelse is_mariadb(Version) of ...` can be used. Look at the JSON test case for example.

qingchuwudi: Version check is a good idea!

Warn (or error) message is necessary (I think). Developers want to know if it is effective and what happened  when the function `reset_connection` is called.

qingchuwudi: Version check is unavailable now.

This driver do not distinguish the difference between **MySQL** and **MariaDB**(Maybe I can try to do the work later).

- mysql server version

    ```bash
    ##myser command:status
    Server version:  5.7.27 MySQL Community Server (GPL)
    ```
    ```erlang
    %%mysql_protocol:nulterm_str/1 recved
    <53,46,55,46,50,55,0,150,0,0,0,75,110,40,2,15,112,35,52,0,255,255,8,2,0,255,193,21,0,...>>
    ```

- MariaDB server version:

    ```bash
    Server version:  5.5.5-10.4.8-MariaDB-1:10.4.8+maria~bionic mariadb.org
    ```
    ```erlang
    %%mysql_protocol:nulterm_str/1 recved
    <<"5.5.5-10.4.8-MariaDB-1:10.4.8+maria~bionic"...>>
    ```

We should use `10.4.8` in MariaDB and `5.7.27` in MySQL, both database **Type** and **Version** will be check.

So I suggest mysql-otp execute ’COM_RESET_CONNECTION‘, and check the response. Only by this way, we can get the real result.

src/mysql_conn.erl
qingchuwudi: Warn (or error) message is helpful. 
Developers want to know if it is effective and what happened when the function reset_connection is called.

src/mysql_conn.erl
```diff
@@ -312,6 +312,20 @@ handle_call({change_user, Username, Password, Options}, From,
...
+handle_call(reset_connection, _From, #state{socket = Socket, sockmod = SockMod} = State) ->
+    setopts(SockMod, Socket, [{active, false}]),
+    Result = mysql_protocol:reset_connnection(SockMod, Socket),
+    setopts(SockMod, Socket, [{active, once}]),
+    State1 = update_state(Result, State),
+    Reply = case Result of
+        #ok{} -> ok;
+        #error{} = E ->
+            %% 'COM_RESET_CONNECTION' is added in MySQL 5.7 and MariaDB 10
+            %% "Unkown command" is returned when MySQL =< 5.6 or MariaDB =< 5.5
```
qingchuwudi: - Removed warning
- more details

test/mysql_tests.erl
zuiderkwast: OK, good investigation! I'm OK with execute COM_RESET_CONNECTION and check the response for this PR. If you want to implement distinguishing between MySQL and MariaDB versions later, it would be awesome.

I would like to solve the failing deadlock test though, so the build passes on travis. I suspect the MySQL version on Travis has changed and it doesn't behave the same as before with deadlocks. I don't have time to investigate this now though.

qingchuwudi:
> Maybe `nulterm_str/1` is only used for the version string, but the idea was to use it for anywhere in the protocol where a nulterminated string is used; maybe for something where else in the future. If it is only for the version, the function name is not good anymore.
> 
> I suggest keeping the original `nulterm_str/1` and instead renaming the function `server_version_to_list/1` to name `parse_server_version` and change the return value to `{Type, Version}`.
> 
> But I think we only need this in the test suites, so why not use query "SELECT @@Version" in the test suite and parse it there? Then, there is no extra overhead and extra bytes stored in the connection itself. I think in applications using this driver, this almost never needed; only for testing I think. (For the connection itself, this is only used to check if the version is >= [5, 0, 0] for `kill_query`.)

Database type is not necessary, and I'll remove the codes.

zuiderkwast:
I have disabled the deadlock test for MySQL 5.7.x, so if you rebase to the latest master, I think it will pass and then I can merge this PR.

qingchuwudi:
> I have disabled the deadlock test for MySQL 5.7.x, so if you rebase to the latest master, I think it will pass and then I can merge this PR.

OK.

qingchuwudi:
> Very good, thank!

You are welcome! I'm going to drop my fork.

PR #139, opened 2019-10-19
Remove error log when tcp_closed

suexcxine:
I met a lot of tcp_closed error message when mysql wait_timeout expires,
I'd rather treat this kind of connection close as normal, not error.
My wait_timeout is 2 hours , when night comes, some connection will stay idle for longger than 2 hours and get closed.
So I got many error alerts......

### DIFF
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -414,7 +414,7 @@ handle_info(ping, #state{socket = Socket, sockmod = SockMod} = State) ->
     setopts(SockMod, Socket, [{active, once}]),
     {noreply, update_state(Ok, State)};
 handle_info({tcp_closed, _Socket}, State) ->
-    stop_server(tcp_closed, State);
+    {stop, normal, State#state{socket = undefined, connection_id = undefined}}; 
 handle_info({tcp_error, _Socket, Reason}, State) ->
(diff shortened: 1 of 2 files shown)


zuiderkwast:
I assume this change fine. If it is restarted (by a supervisor or pool), it would be good to have lazy connection, i.e. connect to the database only when the first query to the database is made.

I'm just thinking about if this is a backwords-incompatible change. For a supervisor, is there a difference? I find in the documentation for supervisor:

> A transient child process is restarted only if it terminates abnormally, that is, with another exit reason than normal, shutdown, or {shutdown,Term}.

It is hard to imagine that anyone is depending on this "transient child process" behavior for a database connection, but I'm keeping this PR open for a few days before I merge it, so other people can comment on it.

suexcxine:
The terminate reason would be 'tcp_closed', it isn't normal or shutdown, so transient child process would be restarted. 
For me, I use poolboy, and the processes are always restarted automatically.

zuiderkwast:
You are aware that the tests are failing?

suexcxine:
I am checking now.

suexcxine:
Done

zuiderkwast:
Accepted! Thanks!

PR #141, opened 2019-11-01
Reconnection support

qingchuwudi:
Because of the problems of network, hardware, and cloud platforms *(Google RDS for example)*, database server is not always available. 
We need to reconncet to database after **several minites** (not  millionseconds or seconds), poolboy and supervisor is not enough at this moment.


qingchuwudi:
close for test cases.

Issue #122, opened 2019-04-24
Lazy connect


juhlig:
In https://github.com/mysql-otp/mysql-otp/pull/105#issuecomment-473861787 I mentioned having problems with `poolboy` when workers (`mysql-otp` connections in my case) fail to start.

I have been discussing this with some people on IRC, their opinion amounts to that a worker should not do more than just starting (ie, not doing the connect yet), and connect (or not) only later, when first used.  This is a recommendation that Fred Hebert also outlines in ["Erlang in Anger"](https://www.erlang-in-anger.com/) (section 2.2).

While I'm actually uncertain about my own opinion on that matter, I think it wouldn't be bad to have some kind of lazy start option, whereby one could decide if `mysql:start_link` should immediately connect, connect only when it is needed for the first time eg on a `query` call (and then stay active), or manually connect with a function like `connect` or something.

What it would mean internally is, give or take, that the connection process just stores the connection information until it actually connects, afterwards it would enter a state much like it currently does right after starting. The functions that do need a real connection (`query`, `prepare`, `execute` etc) would need some auto-connect mechanism. There should be a function to manually connect, like `connect`. And, there should also be a new connect option that decides whether to start immediately (like now), manually (via `connect`), or automatically (when it is first needed, plus `connect`).

zuiderkwast:
This is an interesting topic. I know that this is the preferred way by many. He (Fred Hebert in "Erlang in Anger") writes how you should just return `{error, not_connected}` if the database is down and try to reconnect without crashing the gen_server, the pool and the entire application.

> If, on the other hand, your database is on a remote host, you should expect the connection to fail.  It’s just a reality of distributed systems that things go down. In this case, the only guarantee you can make in the client process is that your client will be able to handle requests, but not that it will communicate to the database. It could return `{error, not_connected}` on all calls during a net split, for example.

Background
-------------

Personally, I had another view when I designed this driver. I didn't want the driver to reconnect, because the state of the connection would be lost. Therefore, I thought it was better to crash. Reconnecting using a backoff timer should be responsibility of a wrapper gen_server outside the connection or the pool or even a process wrapping multiple pools. That's the reasoning behind this design choice.

[...]

PR #142, opened 2019-11-05
Connect modes

juhlig:
In response to latest comments in #122

*  `synchronous` mode (the default) connects to the server as part of the connection processes startup routine. Only when this is done will the call to `start_link` return its `pid`. This is the same as the currently only behavior (with a slight change, see below). If the connection fails, so does the call to `start_link`.
* `asynchronous` mode starts the connection _process_  and immediately returns its `pid`, and will _then_ establish the actual connection. If the connection fails, the connection process will immediately crash again.
* `lazy` mode starts the connection _process_, but does not immediately try to connect. The actual connection will be established later, when it is needed for the first time (for a query, prepare, etc), and then stay connected. If the connection fails, the connection process will crash.

The mentioned change in `synchronous` mode concerns the connect timeout:
* before the on-connect queries and prepares were moved into the connection process, this timeout meant the time taken for connecting the socket and performing the handshake, including a possible ssl upgrade
* after the on-connect queries and prepares were moved into the connection process, this timeout meant all of the before, plus the execution of those queries and prepares
* this PR reverts these changes, as including them in the timeout was an accident that I only now realized

Note that enforcing the connect timeout is much more complicated now that async and lazy connect are involved. Before, it was done by giving a timeout parameter to `start_link`, if connecting and handshaking took too long `gen` would take care of it all.

This is not possible any more. The timeout is now realized by performing the connect and handshake in a spawned process, which hands over the socket to the connection process after it has done that. Meanwhile, the connection process waits for the spawned process to finish and reply (after which it will perform the on-connect things by itself), or for the connect timeout to expire; if that happens, the spawned process is being killed and an error returned, which ultimately results in the connection process exiting.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -26,6 +26,7 @@
 -module(mysql).
 
 -export([start_link/1, stop/1, stop/2,
+         is_connected/1,
          query/2, query/3, query/4, query/5,
          execute/3, execute/4, execute/5,
          prepare/2, prepare/3, unprepare/2,
@@ -61,8 +62,6 @@
                       | {ok, [{column_names(), rows()}, ...]}
                       | {error, server_reason()}.
 
--define(default_connect_timeout, 5000).
-
 -include("exception.hrl").
 
 %% @doc Starts a connection gen_server process and connects to a database. To
@@ -88,6 +87,28 @@
 %%   <dt>`{database, Database}'</dt>
 %%   <dd>The name of the database AKA schema to use. This can be changed later
 %%       using the query `USE <database>'.</dd>
+%%   <dt>`{connect_mode, synchronous | asynchronous | lazy}'</dt>
+%%   <dd>Specifies how and when the connection process should establish a connection
+%%       to the MySQL server.
+%%       <dl>
+%%         <dt>`synchronus' (default)</dt>
+%%         <dd>The connection will be established as part of the connection process'
+%%             start routine, ie the returned connection process will already be
+%%             connected and ready to use, and any on-connect prepares and queries
+%%             will have been executed.</dd>
+%%         <dt>`asynchronous'</dt>
+%%         <dd>The connection process will be started and returned to the caller
+%%             before really establishing a connection to the server and executing
+%%             the on-connect prepares and executes. This will instead be done
+%%             immediately afterwards as the first action of the connection
+%%             process.</dd>
+%%         <dt>`lazy'</dt>
+%%         <dd>Similar to `asynchronous' mode, but an actual connection will be
+%%             esatblished and the on-connect prepares and queries executed only
+%%             when a connection is needed for the first time, eg. to execute a
+%%             query.</dd>
+%%      </dl>
+%%   </dd>
 %%   <dt>`{connect_timeout, Timeout}'</dt>
 %%   <dd>The maximum time to spend for start_link/1.</dd>
 %%   <dt>`{log_warnings, boolean()}'</dt>
@@ -130,6 +151,7 @@
                    {host, inet:socket_address() | inet:hostname()} | {port, integer()} |
                    {user, iodata()} | {password, iodata()} |
                    {database, iodata()} |
+                   {connect_mode, synchronous | asynchronous | lazy} |
                    {connect_timeout, timeout()} |
                    {log_warnings, boolean()} |
                    {keep_alive, boolean() | timeout()} |
@@ -145,13 +167,11 @@
                        {via, Module :: atom(), ViaName :: term()},
          NamedStatements :: [{StatementName :: atom(), Statement :: iodata()}].
 start_link(Options) ->
-    GenSrvOpts = [{timeout, proplists:get_value(connect_timeout, Options,
-                                                ?default_connect_timeout)}],
     case proplists:get_value(name, Options) of
         undefined ->
-            gen_server:start_link(mysql_conn, Options, GenSrvOpts);
+            gen_server:start_link(mysql_conn, Options, []);
         ServerName ->
-            gen_server:start_link(ServerName, mysql_conn, Options, GenSrvOpts)
+            gen_server:start_link(ServerName, mysql_conn, Options, [])
     end.
 
 %% @see stop/2.
@@ -198,6 +218,11 @@ backported_gen_server_stop(Conn, Reason, Timeout) ->
         end
     end.
 
+%% @private
+-spec is_connected(Conn) -> boolean()
+    when Conn :: connection().
+is_connected(Conn) ->
+    gen_server:call(Conn, is_connected).
 
 %% @see query/5.
 -spec query(Conn, Query) -> Result
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -34,6 +34,7 @@
 -define(default_port, 3306).
 -define(default_user, <<>>).
 -define(default_password, <<>>).
+-define(default_connect_timeout, 5000).
 -define(default_query_timeout, infinity).
 -define(default_query_cache_time, 60000). %% for query/3.
 -define(default_ping_timeout, 60000).
@@ -49,10 +50,9 @@
 -include("server_status.hrl").
 
 %% Gen_server state
--record(state, {server_version, connection_id, socket, sockmod, ssl_opts,
-                host, port, user, password, auth_plugin_data, log_warnings,
-                ping_timeout,
-                query_timeout, query_cache_time,
+-record(state, {server_version, connection_id, socket, sockmod, tcp_opts, ssl_opts,
+                host, port, user, password, database, queries, prepares, auth_plugin_data,
+                log_warnings, connect_timeout, ping_timeout, query_timeout, query_cache_time,
                 affected_rows = 0, status = 0, warning_count = 0, insert_id = 0,
                 transaction_levels = [], ping_ref = undefined,
                 stmts = dict:new(), query_cache = empty, cap_found_rows = false}).
@@ -73,14 +73,15 @@ init(Opts) ->
     Database       = proplists:get_value(database, Opts, undefined),
     LogWarn        = proplists:get_value(log_warnings, Opts, true),
     KeepAlive      = proplists:get_value(keep_alive, Opts, false),
-    Timeout        = proplists:get_value(query_timeout, Opts,
+    ConnectTimeout = proplists:get_value(connect_timeout, Opts,
+                                         ?default_connect_timeout),
+    QueryTimeout   = proplists:get_value(query_timeout, Opts,
                                          ?default_query_timeout),
     QueryCacheTime = proplists:get_value(query_cache_time, Opts,
                                          ?default_query_cache_time),
     TcpOpts        = proplists:get_value(tcp_options, Opts, []),
     SetFoundRows   = proplists:get_value(found_rows, Opts, false),
     SSLOpts        = proplists:get_value(ssl, Opts, undefined),
-    SockMod0       = gen_tcp,
 
     Queries        = proplists:get_value(queries, Opts, []),
     Prepares       = proplists:get_value(prepare, Opts, []),
@@ -91,10 +92,66 @@ init(Opts) ->
         N when N > 0 -> N
     end,
 
+    State0 = #state{
+        tcp_opts = TcpOpts,
+        ssl_opts = SSLOpts,
+        host = Host, port = Port,
+        user = User, password = Password,
+        database = Database,
+        queries = Queries, prepares = Prepares,
+        log_warnings = LogWarn,
+        connect_timeout = ConnectTimeout,
+        ping_timeout = PingTimeout,
+        query_timeout = QueryTimeout,
+        query_cache_time = QueryCacheTime,
+        cap_found_rows = (SetFoundRows =:= true)
+    },
+
+    case proplists:get_value(connect_mode, Opts, synchronous) of
+        synchronous ->
+            case connect(State0) of
+                {ok, State1} ->
+                    {ok, State1};
+                {error, Reason} ->
+                    {stop, Reason}
+            end;
+        asynchronous ->
+            gen_server:cast(self(), connect),
+            {ok, State0};
+        lazy ->
+            {ok, State0}
+    end.
+
+connect(#state{connect_timeout = ConnectTimeout} = State) ->
+    MainPid = self(),
+    Pid = spawn_link(
+        fun () ->
+            {ok, State1}=connect_socket(State),
+            case handshake(State1) of
+                {ok, #state{sockmod = SockMod, socket = Socket} = State2} ->
+                    SockMod:controlling_process(Socket, MainPid),
+                    MainPid ! {self(), {ok, State2}};
+                {error, _} = E ->
+                    MainPid ! {self(), E}
+            end
+        end
+    ),
+    receive
+        {Pid, {ok, State3}} ->
+            post_connect(State3);
+        {Pid, {error, _} = E} ->
+            E
+    after ConnectTimeout ->
+        unlink(Pid),
+        exit(Pid, kill),
+        {error, timeout}
+    end.
+
+connect_socket(#state{tcp_opts = TcpOpts, host = Host, port = Port} = State) ->
     %% Connect socket
     SockOpts = [binary, {packet, raw}, {active, false}, {nodelay, true}
                 | TcpOpts],
-    {ok, Socket0} = SockMod0:connect(Host, Port, SockOpts),
+    {ok, Socket} = gen_tcp:connect(Host, Port, SockOpts),
 
     %% If buffer wasn't specifically defined make it at least as
     %% large as recbuf, as suggested by the inet:setopts() docs.
@@ -102,13 +159,18 @@ init(Opts) ->
         true ->
             ok;
         false ->
-            {ok, [{buffer, Buffer}]} = inet:getopts(Socket0, [buffer]),
-            {ok, [{recbuf, Recbuf}]} = inet:getopts(Socket0, [recbuf]),
-            ok = inet:setopts(Socket0,[{buffer, max(Buffer, Recbuf)}])
+            {ok, [{buffer, Buffer}]} = inet:getopts(Socket, [buffer]),
+            {ok, [{recbuf, Recbuf}]} = inet:getopts(Socket, [recbuf]),
+            ok = inet:setopts(Socket, [{buffer, max(Buffer, Recbuf)}])
     end,
 
+    {ok, State#state{socket = Socket}}.
+
+handshake(#state{socket = Socket0, ssl_opts = SSLOpts,
+        user = User, password = Password, database = Database,
+        cap_found_rows = SetFoundRows} = State0) ->
     %% Exchange handshake communication.
-    Result = mysql_protocol:handshake(User, Password, Database, SockMod0, SSLOpts,
+    Result = mysql_protocol:handshake(User, Password, Database, gen_tcp, SSLOpts,
                                       Socket0, SetFoundRows),
     case Result of
         {ok, Handshake, SockMod, Socket} ->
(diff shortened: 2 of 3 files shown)


src/mysql.erl
```diff
@@ -145,13 +144,11 @@
                        {via, Module :: atom(), ViaName :: term()},
          NamedStatements :: [{StatementName :: atom(), Statement :: iodata()}].
 start_link(Options) ->
-    GenSrvOpts = [{timeout, proplists:get_value(connect_timeout, Options,
-                                                ?default_connect_timeout)}],
     case proplists:get_value(name, Options) of
         undefined ->
-            gen_server:start_link(mysql_conn, Options, GenSrvOpts);
+            gen_server:start_link(mysql_conn, Options, []);
```
zuiderkwast: `gen_server` doesn't have a default timeout for `start_link`? I hope not. (For `gen_server:call/2`, the default timeout is 5000.)

src/mysql_conn.erl
```diff
@@ -91,54 +92,109 @@ init(Opts) ->
...
+            do_post_connect(State1);
+        {Pid, {error, _} = E} ->
+            E
+    after ConnectTimeout ->
+        unlink(Pid),
+        exit(Pid, kill),
+        {error, timeout}
+    end.
+
+do_connect1(#state{tcp_opts = TcpOpts, host = Host, port = Port} = State) ->
```
zuiderkwast: The names do_connect, do_connect1, do_connect2 could be better, e.g. by including words like handshake, depending on what they do. Comments (or edoc) can also be used to describe them if the names are not clear enough.

src/mysql_conn.erl
```diff
@@ -91,54 +92,109 @@ init(Opts) ->
...
+        lazy ->
+            {ok, State0}
+    end.
+
+do_connect(#state{connect_timeout = ConnectTimeout} = State) ->
+    Self = self(),
+    Pid = spawn_link(
+        fun () ->
+            case do_connect1(State) of
+                {ok, #state{sockmod = SockMod, socket = Socket} = State1} ->
```
zuiderkwast: Is this spawned just to be able to set a timeout using receive after? It would be more clean connect in the same process and just use recv timeouts on tcp/ssl. The timeout value doesn't need to be a hard limit IMO.

src/mysql_conn.erl
```diff
@@ -91,54 +92,109 @@ init(Opts) ->
...
+        exit(Pid, kill),
+        {error, timeout}
+    end.
+
+do_connect1(#state{tcp_opts = TcpOpts, host = Host, port = Port} = State) ->
     %% Connect socket
     SockOpts = [binary, {packet, raw}, {active, false}, {nodelay, true}
                 | TcpOpts],
-    {ok, Socket0} = SockMod0:connect(Host, Port, SockOpts),
+    {ok, Socket} = gen_tcp:connect(Host, Port, SockOpts),
```
zuiderkwast: Here `gen_tcp:connect(Address, Port, Options, Timeout)` can be used, instead of the extra process for timeout handling.

src/mysql_conn.erl
```diff
@@ -437,8 +512,8 @@ code_change(_OldVsn, _State, _Extra) ->
 
 %% --- Helpers ---
 
-%% @doc Executes a prepared statement and returns {reply, Reply, NextState}.
-execute_stmt(Stmt, Args, FilterMap, Timeout,
+%% @doc Executes a prepared statement and returns {Reply, NewState}.
+do_execute_stmt(Stmt, Args, FilterMap, Timeout,
```
zuiderkwast: Why the "do"? Is it intended as a prefix for all internal functions? I don't like it very much as it makes the names clumsier. :-)

zuiderkwast:
Btw, tell me if you disagree about some comment. I can change my mind. :-)

src/mysql.erl
juhlig: Oh, right, I didn't think of that. So I have just been looking into what `gen_server` and `gen` do there:

[`gen_server:start_link/3,4`](https://github.com/erlang/otp/blob/36fdda7ff23fe8d163283287d950ae903de359d1/lib/stdlib/src/gen_server.erl#L179) call [`gen:start/5,6`](https://github.com/erlang/otp/blob/36fdda7ff23fe8d163283287d950ae903de359d1/lib/stdlib/src/gen.erl#L75), which in turn calls [`gen:do_spawn/5,6`](https://github.com/erlang/otp/blob/36fdda7ff23fe8d163283287d950ae903de359d1/lib/stdlib/src/gen.erl#L92), which then uses [`gen:timeout/1`](https://github.com/erlang/otp/blob/36fdda7ff23fe8d163283287d950ae903de359d1/lib/stdlib/src/gen.erl#L362) to retrieve the timeout from the options, defaulting to `infinity` if it is absent.

src/mysql_conn.erl
juhlig: Yes, just for that reason.

I had built into connect etc in the beginning, but found that I had to push that timeout around very often: `gen_tcp:connect`, `ssl:connect` (where it is currently hardcoded to `5000`), and also to all the `send` and `recv` calls that are involved in the various steps of the handshake to be consistent. And actually, I'd have to keep track and reduce it with every step taken to stay close to the overall timeout, otherwise the overall time taken might add up and get quite far from the desired timeout value. Also, there needs to be timeout handling put in place like everywhere, eg without a timeout, `gen_tcp:connect` will always return `{ok, Socket}` sooner or later (unless there are other errors, of course), while with a timeout it may also return `{error, timeout}` or something, etc, same for all the `send` and `recv` calls.

For that reason, I switched to connecting in a separate process, which can run much as it does now, but gets killed from the outside if it takes too long for the overall process.

src/mysql_conn.erl
juhlig: In fact, I had it like this before, forgot why I switched over to 1, 2, ... =^^=

src/mysql_conn.erl
juhlig: It's a habit, I guess =^^= I often do it like this in other projects, at least until the interface has settled down a little. This way, I can always export nicely named functions and don't have to think up new ones or rename things in case they are already taken by something internal.

In this case, IIRC, I did it to be more consistent with `do_query` (`query` seems to be a reserved word, though I have no idea what it is ever for).

juhlig:
I didn't notice the emoji until you mentioned it just now =^^=

As for tests, I'll do them once we have ironed out the actual thing. Just wanted to let you take a look at it and hear your opinions before I go further, lest you say sth like "this is _totally_ not how I want it" :)

Anyway, I had been playing with the tests in `mysql_tests` a bit already. Most of the tests there could be run in all three modes, actually, but I have found no nice easy way to do that with `eunit`, at least the test output is always kinda funky =^^= I usually prefer `ct` in my projects, so am not all that familiar with everything you can do with `eunit` and how.

juhlig:
Latest commit introduces `is_connected/1` and four test cases (not sure two are necessary for `asynchronous`, let me know what you think :))

src/mysql.erl
zuiderkwast: Great.

src/mysql_conn.erl
zuiderkwast: Fine then!

So I guess the ownership of the socket gets transferred when the connecting process is done? I mean, for `{active, once}` later, messages like `tcp_closed` have to be delivered to the connection process, not to the connecting helper-process which isn't alive anymore.

src/mysql_conn.erl
zuiderkwast: Query was a reserved word in Erlang until Luic removed the reservation in a PR some years ago. It wasn't used for anything. This was before I started this project and `mysql:query/2,3,4,5` exist indeed. :-)

Anyway, the only functions exported in this module are the gen_server behaviour callbacks, so no need to prefix with "do" in general, so at least don't rename existing functions like this one.

src/mysql_conn.erl
zuiderkwast: Ah, but please think about the names and possibly split them further.

Any names like connection, authenticate, do_handshake, pre_connect, etc. are better IMO. With numbers you can't split a function in two without changing all the numbers. And "connecting" is actually just the tcp part. All the rest is something else, if you think about it.

src/mysql_conn.erl
```diff
@@ -91,54 +92,109 @@ init(Opts) ->
...
+    end.
+
+do_connect(#state{connect_timeout = ConnectTimeout} = State) ->
+    Self = self(),
+    Pid = spawn_link(
+        fun () ->
+            case do_connect1(State) of
+                {ok, #state{sockmod = SockMod, socket = Socket} = State1} ->
+                    SockMod:controlling_process(Socket, Self),
+                    Self ! {self(), {ok, State1}};
```
zuiderkwast: The Self variable could be called MainPid or something. Self is not self everywhere, so it's not a very good name.

src/mysql.erl
```diff
@@ -130,6 +129,7 @@
                    {host, inet:socket_address() | inet:hostname()} | {port, integer()} |
                    {user, iodata()} | {password, iodata()} |
                    {database, iodata()} |
+                   {connect_mode, synchronous | asynchronous | lazy} |
```
zuiderkwast: The option is missing in the edoc above.

src/mysql.erl
```diff
@@ -198,6 +196,10 @@ backported_gen_server_stop(Conn, Reason, Timeout) ->
         end
     end.
 
+-spec is_connected(Conn) -> boolean()
```
zuiderkwast: One line of edoc wouldn't hurt, e.g.

```Erlang
%% @doc Always returns true, except for connections started with
%% connect mode `lazy' or `asynchronous' where false can be retuned
%% before a connection has been established.
```

juhlig: This is not entirely accurate. Even in asynchronous mode, it will always return `true`: The connection process spawns the temporary process that does the actual connect to the server and waits for completion in a `receive`. Thus, it won't react to any message sent to it (like with `gen_server:call`) while it waits at this point, and when it is finally able to react, it also means the connection has been established.

Now that I think about it, what _may_ happen is that `is_connected/1` times out, depending on how long the connect takes in asynchronous mode, or if the connection process is busy executing a query or something in all modes. So I better provide a `_is_connected/2` which takes a timeout, yes?

src/mysql_conn.erl
juhlig: Another bad habit of mine, will change :)

src/mysql_conn.erl
juhlig: > And "connecting" is actually just the tcp part. All the rest is something else, if you think about it.

I know, but what would you call the overall process of connecting, authentication, ssl upgrade, etc? The one function that wraps all that, I mean. Nothing comes to mind for me =^^=

src/mysql.erl
zuiderkwast: Nah, not worth it, is it? Just put a `%% @private` so it doesn't show up in the documentation.

src/mysql_conn.erl
zuiderkwast: Hm... ok, I will think about it. Perhaps I can come up with something.

src/mysql_conn.erl
juhlig: Strange... I googled a bit, what I found says `query` was removed in OTP 17. But the travis tests for mysql-otp also use R16B03-1 and pass despite `mysql:query` being in there, so I guess it was removed even earlier.

src/mysql_conn.erl
zuiderkwast: When the functions are chained using tail calls, it's hard to name them and dependency-wise they all depend on each other. Thus, I suggest not chaining them using tail calls.

Instead call them one by one from `do_connect` (which I think can simply be called `connect`). Rename `do_connect1` => `connect_tcp` and let it return the updated state with the socket set. Then rename `do_connect2` => `do_handshake` and don't call `do_post_connect`. `do_post_connect` can also be called directly from `connect` (but a case is needed for reporting handshake errors properly).

This puts some more code in the already deeply indented argument to spawn_link. This can be moved to it's own function, which can take the name `do_connect` (which in my world indicates that it's a helper to the `connect` function).

```
connect     do_connect
   |           |
   |---spawn-->|     
   |           |<-------> connect_tcp
   |           |<-------> do_handshake
   |           |<-------> do_post_connect
   |<-receive--|
```
instead of
```
do_connect   fun()
    |          |
    |--spawn-->|
    |          |--> do_connect1 -> do_connect2 -> do_post_connect
    |          |<----(return)--------------------------|
    |<-receive-|
```

src/mysql_conn.erl
```diff
@@ -91,54 +92,109 @@ init(Opts) ->
...
-            end;
+                           status = Status},
+            do_post_connect(State1);
         #error{} = E ->
-            {stop, error_to_reason(E)}
+            {error, error_to_reason(E)}
+    end.
+
+do_post_connect(#state{queries = Queries, prepares = Prepares} = State) ->
+    case execute_on_connect(Queries, Prepares, State) of
```
zuiderkwast: To not add to technical dept, this can be used in `change_user` as well. It has almost the same logic.

src/mysql_conn.erl
zuiderkwast: Look here: https://github.com/erlang/otp/pull/1525

src/mysql_conn.erl
juhlig: Changed as suggested in latest commit.

src/mysql_conn.erl
juhlig: Hm, I made some changes just before receiving this comment. I renamed `do_connect` to `connect`, `do_connect1` to `connect_socket`, `do_connect2` to `handshake`, and `do_post_connect` to `post_connect`.

Other than you depict, `post_connect` is not done in the spawned process but after the receive, and is thus not affected by the connect timeout. This is for two reasons: For one, I'd say it is not part of the actual connection procedure, but afterwards, and thus should not be affected by the connect timeout; for another, this has been the behavior before the on-connect prepares and queries were moved into the connection process (IIRC, by my last PR before this one), putting the on-connect stuff under the connect timeout was an accident.

src/mysql_conn.erl
juhlig: Ah, so looks like it was done in R16B02. Could have asked Loïc himself, come to think of it =^^=

Anyway, I renamed `do_execute_stmt` back to `execute_stmt`, and did the same for `do_query` and `do_named_prepare` while I was at it.

src/mysql.erl
juhlig: Added in latest commit, explanation has become quite verbose but should cover everything a user may need to know.

src/mysql.erl
juhlig: Ok, done in latest commit.

src/mysql_conn.erl
juhlig: Hm, I don't know if it achieves much (unless I missed the point), but did as suggested.

juhlig: Looking back at it all, I get the feeling that `mysql_conn` has overall become a bit messy over time. When this one is through, I may look into pulling some things more straight, I guess :)

src/mysql.erl
zuiderkwast: Aha :) i meant @private to hide the function entirely, so we don't need to document it. Not great if it's i correct i guess. Not a big deal though.

zuiderkwast: Aha :) i meant @private to hide the function entirely, so we don't need to document it. Not great if it's i correct i guess. Not a big deal though.

src/mysql_conn.erl
```diff
@@ -300,8 +368,10 @@ handle_call({change_user, Username, Password, Options}, From,
     State2 = State1#state{query_cache = empty, stmts = dict:new()},
     case Result of
         #ok{} ->
-            State3 = State2#state{user = Username, password = Password},
-            case execute_on_connect(Queries, Prepares, State3) of
+            State3 = State2#state{user = Username, password = Password,
+                                  database=Database, queries=Queries,
+                                  prepares=Prepares},
```
zuiderkwast: What's happening here, setting database, queries and prepares?

The benefit of using post_connect here would be that we don't need to fetch queries and prepares from the state, since post_connect does that.

src/mysql.erl
```diff
@@ -88,6 +87,28 @@
 %%   <dt>`{database, Database}'</dt>
 %%   <dd>The name of the database AKA schema to use. This can be changed later
 %%       using the query `USE <database>'.</dd>
+%%   <dt>`{connect_mode, synchronous | asynchronous | lazy}'</dt>
+%%   <dd>Specifies how and when the connection process should establish a connection
+%%       to the MySQL server.
+%%       <dl>
+%%         <dt>`synchronus' (default)</dt>
+%%         <dd>The connection will be established as part of the connection process'
```
zuiderkwast: Awesome documentation!

src/mysql.erl
zuiderkwast: Aha :) i meant @private to hide the function entirely, so we don't need to document it. Not great if it's i correct i guess. Not a big deal though.

src/mysql_conn.erl
zuiderkwast: Really good! I agree post_connect should be after the receive.

src/mysql.erl
juhlig: Ah, ok, I'll remove the documentation again then.

src/mysql_conn.erl
juhlig: So this is ok now, yes?

Btw, this showed me an omission I had made earlier, I forgot to put the queries and prepares from the change user call into the state. Not a big deal actually, as they are never used again after this, but still.

src/mysql.erl
juhlig: Thanks :)

juhlig:
The latest commit should round this up finally. As far as I can see, all requested changes have been accounted for, right?

Anyway, you may want to look this over again before merging. The test suites pass, but there may still be bugs hidden somewhere. For example, while doing one of the requested changes, I noticed that the on-connect queries and prepares were in fact done twice.

src/mysql_conn.erl
zuiderkwast: That would be nice!

src/mysql_conn.erl
zuiderkwast: Ah, good. I didn't remember that change user contains `{change_user, Username, Password, Options}`. I imagined that the same prepares and queries as in the initial connect would be used after change user. I see now.

Also good that it gets saved, for consistency. They're not needed, but at least the old values are erased from the state in this way.

src/mysql_conn.erl
```diff
@@ -408,7 +487,8 @@ handle_info(query_cache, #state{query_cache = Cache,
     {noreply, State#state{query_cache = Cache1}};
 handle_info({'DOWN', _MRef, _, Pid, _Info}, State) ->
     stop_server({application_process_died, Pid}, State);
-handle_info(ping, #state{socket = Socket, sockmod = SockMod} = State) ->
+handle_info(ping, #state{socket = Socket, sockmod = SockMod} = State)
+  when Socket =/= undefined ->
```
zuiderkwast: Is this check needed? Ping is scheduled in post_connect, so we should always be connected here. If we're not connected, we have the problem that the ping is not re-scheduled...

juhlig: Hm, looks like a leftover from a previous iteration I did... You're right, it is not needed, I'll remove it.

juhlig: Removed in latest commit.

src/mysql_conn.erl
juhlig: Using the same prepares and queries would probably be undesirable, maybe even dangerous, I think =^^= Certainly so if the connections are used in a pool.

Looking at the state record, I see that it carries around a lot of stuff that is not needed after connecting, like the tcp and ssl options, username, password, queries, prepares, ... When I get around to do a cleanup as mentioned above, I think it might be feasible (at a cursory glance) to have in fact _two_ state records, one that carries what is needed for connecting, one that carries what is needed after connecting, start with the not connected state, and replace the gen_server state when "switching" from not connected to connected. I have been doing this a few times in own projects, though often its not worth it, and not sure it is possible here. I'll investigate :)

zuiderkwast:
Great! I edited the commit message a bit, to include the different modes in the first line.

src/mysql_conn.erl
zuiderkwast: Two states sounds good! Perhaps the state before connecting can just be the list of options received in `init/1`. Then, the check if we're connected is just an `is_list(State)`. I don't know if it's strait-forward but it's an idea at least.

juhlig: I started experimenting with this, but turns out we can't do it like this... Problem is that we need host, port, user, password etc in kill_query. So, there isn't much what could be removed from the state record. Not worth the effort I think :\

zuiderkwast: OK then. Thanks for investigating this though.

PR #143, opened 2019-11-15
Fix backslash encode

suexcxine:
I met syntax error when generating sql with mysql_encode:encode,
Only single quotes is being encoded now,
So backslashes also need to be encoded.

### DIFF
--- a/src/mysql_encode.erl
+++ b/src/mysql_encode.erl
@@ -17,7 +17,7 @@ encode(Float) when is_float(Float) ->
     %% "floats are printed accurately as the shortest, correctly rounded string"
     io_lib:format("~w", [Float]);
 encode(Bin) when is_binary(Bin) ->
-    Escaped = binary:replace(Bin, <<"'">>, <<"''">>, [global]),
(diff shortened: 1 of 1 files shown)


zuiderkwast:
Good. Don't use `mysql_encode`. It's an internal ondocumented module. Use `mysql:encode/2` which is documented.

In the SQL standard, only single quote is encoded by two single quotes.

Backslash escape is MySQL specific and it only works when certain option is enabled in MySQL. If standard SQL mode is enabled, backslash escape cannot be used. The connection has some flag from the server which is used to check if backslashes should be encoded or not, so it's automatic. Just don't use the internals directly. :-)

zuiderkwast:
https://mysql-otp.github.io/mysql-otp/mysql.html

suexcxine:
Thank you for your detailed explaination.
There is a gen_server:call in mysql:encode, which would be a performance penalty(I am afraid).
I set NO_BACKSLASH_ESCAPES globally and problem solved.

zuiderkwast:
Sure, as long as you know what you're doing 👍

zuiderkwast:
Btw, a gen_server call is a small performance hit, but in relation to making a query to a database over the network, it's nothing.

Issue #147, opened 2019-12-04
Prepending to TCP options list in connect_socket


juhlig:
https://github.com/mysql-otp/mysql-otp/blob/08b6eabfe41e980613300092763335efb8c6fa09/src/mysql_conn.erl#L152-L154

Adding something to the socket options this way is "unsafe". If one of those options is present in `TcpOpts`, they will be used, as the last occurence in the list is the one that gets used by `gen_tcp:connect`. For `mode` (or the same via `binary` / `list`), `packet` and `active`, this will result in an error if it is not `binary`, `0` (or `raw`), and `false`, respectively, while for `nodelay` what is in `TcpOpts` just overrides `true`.

Not sure if this should be changed and to what behavior. After all, options that would make a connection unusable when invalid (`mode` not `binary`, `packet` not `0` or `raw`, `active` not `false`) _do_ result in an error, and with `nodelay` set to `false` the connection would still work. What do you think?

zuiderkwast:
Allowing TcpOpts is good for anyone who knows what they're doing and need to tweak the connection. For those who actively try to mess things up, they can try to do so and then revert to the defaults when they realize it doesn't work, so I didn't put any effort in preventing this.

If you want, you can check that neither of `packet`, `active` nor `binary` is defined in TcpOpts and raise an `error(badarg)` if they are, in `start_link/1`. We can consider `nodelay` just as a default which can be overridden if anyone wants to do that.

Does this seem reasonable?

juhlig:
Instead of going for some big checking gymnastics, we could just _append_ "our" parameters to the given TCP options list, instead of _prepending_. This should ensure they are what is used regardless of what the user gave, notwithstanding any changes in how OTP handles the options list, which I think is unlikely.

zuiderkwast:
OK! Prepend nodelay and append the others then?

juhlig:
Ah, good idea :) Will see to it tomorrow :)

PR #148, opened 2019-12-06
Override mode, packet and active TCP options

juhlig:
#147 

The options `mode` (including `binary`/`list`), `packet` and `active` that could be given in the TCP options by the user are overridden with correct values `binary`, `0` and `false`, respectively.

The `nodelay` option defaults to `true`, but may be overridden by the user if s/he so wishes.

### DIFF
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -149,8 +149,7 @@ connect(#state{connect_timeout = ConnectTimeout} = State) ->
 
 connect_socket(#state{tcp_opts = TcpOpts, host = Host, port = Port} = State) ->
     %% Connect socket
-    SockOpts = [binary, {packet, raw}, {active, false}, {nodelay, true}
-                | TcpOpts],
+    SockOpts = sanitize_tcp_opts(TcpOpts),
     {ok, Socket} = gen_tcp:connect(Host, Port, SockOpts),
 
     %% If buffer wasn't specifically defined make it at least as
@@ -165,6 +164,24 @@ connect_socket(#state{tcp_opts = TcpOpts, host = Host, port = Port} = State) ->
     end,
 
     {ok, State#state{socket = Socket}}.
+
+sanitize_tcp_opts(TcpOpts0) ->
+    TcpOpts1 = lists:filter(
+        fun
+            ({mode, _}) -> false;
+            (binary) -> false;
+            (list) -> false;
+            ({packet, _}) -> false;
+            ({active, _}) -> false;
+            (_) -> true
+        end,
+        TcpOpts0
+    ),
+    TcpOpts2 = case lists:keymember(nodelay, 1, TcpOpts1) of
+        true -> TcpOpts1;
+        false -> [{nodelay, true} | TcpOpts1]
+    end,
+    [binary, {packet, raw}, {active, false} | TcpOpts2].
 
 handshake(#state{socket = Socket0, ssl_opts = SSLOpts,
         user = User, password = Password, database = Database,


zuiderkwast:
Actually, are you sure that the last specified option has precedence? I can't find it in the documentation for gen_tcp. It looks like a proplist and with `proplists:get_value/2,3`, the first match is used.

juhlig:
Let's say I'm _pretty sure_ :) For one, it can't be a `proplist`, as for instance the `raw` option can occur multiple times and is not a 2-tuple. I rummaged through the `gen_tcp` and `inet` modules, and from what I can see, the options list is being iterated front to end to transform it into a map, which explains why the last occurrence of an option takes precedence.

juhlig:
Oh, and I tried some `gen_tcp:connect` with repeated options of different values by hand, of course ;)

zuiderkwast:
> Let's say I'm _pretty sure_ :) For one, it can't be a `proplist`, as for instance the `raw` option can occur multiple times and is not a 2-tuple. I rummaged through the `gen_tcp` and `inet` modules, and from what I can see, the options list is being iterated front to end to transform it into a map, which explains why the last occurrence of an option takes precedence.

I believe you. But in OTP < 19 there are no maps.

Also `lists:keyfind/3` returns the first match.

What I mean is that it's not documented and that means it may vary between versions and possibly change in the future without notice. The safe way would be to ensure that there are no ambiguities.

```Erlang
filter_opt(list)        -> false;
filter_opt({mode, _})   -> false;
filter_opt({packet, _}) -> false;
filter_opt({active, _}) -> false;
filter_opt(_)           -> true.

connect_socket(#state{tcp_opts = TcpOpts, host = Host, port = Port} = State) ->
    SockOpts = [binary, {packet, raw}, {active, false}, {nodelay, true}
                | lists:filter(fun filter_opt/1, TcpOpts)],
    ...
```

If we care to prevent the user from messing up, that is.  If we don't, there is no reason to change anything here IMO.

juhlig:
> I believe you. But in OTP < 19 there are no maps.

[OTP 17](http://erlang.org/documentation/doc-6.0/doc/reference_manual/maps.html), actually ;) But anyway, I just so noticed that I have misread the part in `inet` where this is done, and they use a _record_ to store things, not a map.

> What I mean is that it's not documented and that means it may vary between versions and possibly change in the future without notice. The safe way would be to ensure that there are no ambiguities.

Ah, ok, I get the point :) In that case, I have to say that, while I can confirm that the appending approach works all the way back to R16B03, the fact that it does is only circumstantial (not intended), and may indeed be subject to changes without notice. While I think that in itself unlikely, with the reworking of `gen_tcp` to use `sockets` at the horizon, who knows what else may change in its wake.

> If we care to prevent the user from messing up, that is. If we don't, there is no reason to change anything here IMO.

So, what do you say, _do_ we care? :) The change was a no-brainer while it only involved putting "our" params at the end, but if we have to resort to active filtering, I'd say we save ourselves the trouble and let the user shoot his own foot if that's what s/he wants to do?

zuiderkwast:
That's a thorough investigation! I hope you enjoyed it. :-)

I don't think active filtering is much more complex than the prepend+append approach. It's purpose is clear without any code comment. If the list to filter is empty, it's a no-op anyway.

Anyway, it's not important to me so I let you decide. Any of the 3 options is OK. I have simply given my input.

juhlig:
> That's a thorough investigation! I hope you enjoyed it. :-)

Actually, I did :) To be fair, I didn't check with _every_ version between R16B03 and current, only the latest of each major release respectively.

> Anyway, it's not important to me so I let you decide. Any of the 3 options is OK. I have simply given my input.

Ok, with my last commit, I made it so that...
1. any occurrence of `{mode, _}`, `list`, `binary`, `{packet, _}` and `{active, _}` are removed from the options
2. if `{nodelay, _}` is _not_ in the options, `{nodelay, true}` is inserted into the options
3. `{binary`, `{packet, raw}` and `{active, false}` is inserted into the options

zuiderkwast:
OK, good! Merged and end of discussion. :-)

Issue #12, opened 2014-11-29
Log slow and unindexed queries (options)


zuiderkwast:
In each result we get some status bits from the server with the meanings "query was slow", "no index was used", "no good index was used". When this happens, log the query using `error_logger:warning_msg("MySQL query was slow: ~s", [Query])`, etc.

Options:
- `{log_slow_queries, boolean()}`, default `false`
- <strike>`{log_no_index_used, boolean()}` default false? Performing queries with no index is always bad.</strike>
- <strike>`{log_no_good_index_used, boolean()}` default false?</strike>

I think we should log only slow queries and add the "index used" info to the "slow query" log entry. No index querys that are not slow are not logged. Examples:

```
"MySQL query (with no index) was slow: SELECT ..."
"MySQL query (with no good index) was slow: SELECT ..."
"MySQL query was slow: SELECT ..."
```

[Edited 20 March 2018]

m13m:
Ping @zuiderkwast  can I take a shot. From the C/S protocol page 

Do we have to warn only for the following status flag?

`SERVER_STATUS_NO_GOOD_INDEX_USED`
`SERVER_STATUS_NO_INDEX_USED`
`SERVER_QUERY_WAS_SLOW`

Any suggestion before I start working on this?

juhlig:
I wanted to give this a try, but really I could not find an (easy) way to cause the slow query flag to be set in the server status. I tried with MySQL 5.7 and `SELECT SLEEP(11)`, then got the connection process' state with `sys:get_state` to see what's in the status field, but there is always just a 2 (meaning autocommit). On the other hand, the query _does_ show up in the slow query log. Am I missing something here?

zuiderkwast:
I don't know an easy way either. If it shows up in the slow query log but not in the client, I don't know what's happening. If there is a config option, I don't know about it.

The way I can think of is to populate the database with random data and no index or no good index, possibly with many workers to get some load on the database as well. There is no reason to run such a test by default on all OTP versions. A more sophisticated travis file would be necessary for that.

If we can reproduce it and capture the traffic (using tcpdump or ngrep), we can add it as a protocol unit test, i.e. mock the server (see `mysql_protocol_tests`). If we can't even make the server produce these warnings, we can write a server response by hand based on the protocol documentation.

zuiderkwast:
Btw, it would be nice to use log using logger where available, with error_logger as a fallback, for this and the existing logging. This could be a separate issue.

[...]

PR #149, opened 2019-12-11
Examine final EOF packet following resultset rows

juhlig:
As I found out when working on #12, the final `EOF` packet following the rows of a result set never gets examined, thereby some status flags may get dropped.


zuiderkwast:
Awesome!

PR #150, opened 2019-12-11
Log slow queries

juhlig:
#12

This PR is based on #149.

The option to enable slow query logging is `{log_slow_queries, boolean()}`, and defaults to `false` (by intuition, can change that to `true` if you think that's better).

I also put a number in the generated log. This is for multi statements (for single statements, it is always 1), as though we can get the status and thus the slow query flag for each individual resultset and/or ok, it is not possible to get the individual query, we only have the entire query string. The number specifies which one of the individual queries was found to be slow.

As for the test, while I found the specific MySQL version where the `SERVER_QUERY_WAS_SLOW` flag appeared to be 5.5.8 and so the test is skipped on earlier MySQL versions, I couldn't really find the respective MariaDB version, only 10.0.21 being mentioned in a related forum post, so I nailed MariaDB to this version.

### DIFF
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -52,7 +52,8 @@
 %% Gen_server state
 -record(state, {server_version, connection_id, socket, sockmod, tcp_opts, ssl_opts,
                 host, port, user, password, database, queries, prepares, auth_plugin_data,
-                log_warnings, connect_timeout, ping_timeout, query_timeout, query_cache_time,
+                log_warnings, log_slow_queries,
+                connect_timeout, ping_timeout, query_timeout, query_cache_time,
                 affected_rows = 0, status = 0, warning_count = 0, insert_id = 0,
                 transaction_levels = [], ping_ref = undefined,
                 stmts = dict:new(), query_cache = empty, cap_found_rows = false}).
@@ -72,6 +73,7 @@ init(Opts) ->
     Password       = proplists:get_value(password, Opts, ?default_password),
     Database       = proplists:get_value(database, Opts, undefined),
     LogWarn        = proplists:get_value(log_warnings, Opts, true),
+    LogSlow        = proplists:get_value(log_slow_queries, Opts, false),
     KeepAlive      = proplists:get_value(keep_alive, Opts, false),
     ConnectTimeout = proplists:get_value(connect_timeout, Opts,
                                          ?default_connect_timeout),
@@ -99,7 +101,7 @@ init(Opts) ->
         user = User, password = Password,
         database = Database,
         queries = Queries, prepares = Prepares,
-        log_warnings = LogWarn,
+        log_warnings = LogWarn, log_slow_queries = LogSlow,
         connect_timeout = ConnectTimeout,
         ping_timeout = PingTimeout,
         query_timeout = QueryTimeout,
@@ -537,7 +539,7 @@ execute_stmt(Stmt, Args, FilterMap, Timeout,
     State1 = lists:foldl(fun update_state/2, State, Recs),
     State1#state.warning_count > 0 andalso State1#state.log_warnings
         andalso log_warnings(State1, Stmt#prepared.orig_query),
-    handle_query_call_result(Recs, Stmt#prepared.orig_query, State1, []).
+    handle_query_call_result(Recs, Stmt#prepared.orig_query, State1).
 
 %% @doc Produces a tuple to return as an error reason.
 -spec error_to_reason(#error{}) -> mysql:server_reason().
@@ -565,10 +567,10 @@ update_state(Rec, State) ->
 
 %% @doc executes an unparameterized query and returns {Reply, NewState}.
 query(Query, FilterMap, default_timeout,
-            #state{query_timeout = DefaultTimeout} = State) ->
+      #state{query_timeout = DefaultTimeout} = State) ->
     query(Query, FilterMap, DefaultTimeout, State);
 query(Query, FilterMap, Timeout,
-            #state{sockmod = SockMod, socket = Socket} = State) ->
+      #state{sockmod = SockMod, socket = Socket} = State) ->
     setopts(SockMod, Socket, [{active, false}]),
     Result = mysql_protocol:query(Query, SockMod, Socket, FilterMap, Timeout),
     {ok, Recs} = case Result of
@@ -587,7 +589,7 @@ query(Query, FilterMap, Timeout,
     State1 = lists:foldl(fun update_state/2, State, Recs),
     State1#state.warning_count > 0 andalso State1#state.log_warnings
         andalso log_warnings(State1, Query),
-    handle_query_call_result(Recs, Query, State1, []).
+    handle_query_call_result(Recs, Query, State1).
 
 %% @doc Prepares a named query and returns {{ok, Name}, NewState} or
 %% {{error, Reason}, NewState}.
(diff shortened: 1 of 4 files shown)


juhlig:
Oh, forgot the docs... Will do it tomorrow 😊

juhlig:
`log_slow_queries` option documented. Also fixed two typos in the documentation of the `connect_mode` option on the way.

zuiderkwast:
Seems great! I will look tomorrow, or any day as soon.

test/mysql_tests.erl
```diff
@@ -373,6 +373,44 @@ log_warnings_test() ->
...
+                throw({mysql, version_too_small});
+            _ ->
+                ok
+        end
+    of _ ->
+        ok = mysql:query(Pid, "SET long_query_time = 0.1"),
+        Query = "SELECT SLEEP(0.2); " %% #1 -> slow
+                "SELECT 1; "          %% #2 -> not slow
+                "SELECT SLEEP(0.2); " %% #3 -> slow
+                "SELECT 1",           %% #4 -> not slow
```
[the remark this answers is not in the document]
zuiderkwast: What if some of the queries don't return any result sets, e.g. `INSERT` or `UPDATE`? Then, is there a risk that the numbers of the queries don't match? If yes, a note in the documentation would be nice.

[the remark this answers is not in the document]
juhlig: Hm, I guess you're right, didn't think of that. Will check again on monday, can't do much now, I'm only on my mobile now 😩

test/mysql_tests.erl
```diff
@@ -373,6 +373,44 @@ log_warnings_test() ->
                  " in INSERT INTO foo () VALUES ()\n", Log3),
     mysql:stop(Pid).
 
+log_slow_queries_test() ->
+    {ok, Pid} = mysql:start_link([{user, ?user}, {password, ?password},
+                                  {log_warnings, false}, {log_slow_queries, true}]),
+    Version = db_version_string(Pid),
+    try
+        Version1 = parse_db_version(Version),
```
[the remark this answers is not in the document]
juhlig: Ok, will do :)

src/mysql_conn.erl
```diff
@@ -667,6 +671,21 @@ log_warnings(#state{socket = Socket, sockmod = SockMod}, Query) ->
...
+  when S band ?SERVER_QUERY_WAS_SLOW /= 0 ->
+    IndexHint = if
+        S band ?SERVER_STATUS_NO_GOOD_INDEX_USED /= 0 ->
+            " (with no good index)";
+        S band ?SERVER_STATUS_NO_INDEX_USED /= 0 ->
+            " (with no index)";
+        true ->
+            ""
+    end,
+    error_logger:warning_msg("MySQL query #~b~s was slow: ~s~n", [N, IndexHint, Query]);
```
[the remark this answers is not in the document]
juhlig: Done in last commit. Slightly added complexity, in a nutshell, `handle_query_call_result` checks if a single or a multi-result returned, and passes that information along, and `maybe_log_slow_query` decides if it needs to include query numbering in the log output or not.

test/mysql_tests.erl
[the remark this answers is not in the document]
juhlig: Done in last commit.

test/mysql_tests.erl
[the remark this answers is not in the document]
juhlig: Turns out the numbering will be correct even if some of the queries don't return result sets. But I added such a query to the test case for good measure :)

juhlig:
The test failure is due to a race condition in that test. This has been addressed in #151.

src/mysql_conn.erl
```diff
@@ -616,13 +618,18 @@ named_prepare(Name, Query, State) ->
 
 %% @doc Transforms result sets into a structure appropriate to be returned
 %% to the client.
-handle_query_call_result([], _Query, State, []) ->
+handle_query_call_result([_] = Recs, Query, State) ->
+    handle_query_call_result(Recs, single_result, 1, Query, State, []);
```
zuiderkwast: Awesome! Just an idea: Can these two can be combined into one: If it's an integer it means it's a multi-result and for single-results we can use an atom, such as for example `ResultNum :: pos_integer() | not_applicable`. They only have a meaning together anyway and some combinations are impossible (`single_result` with N > 1).

I realize we need an extra case somewhere for the recursion, but on the other hand there will be one parameter less in other places... So you can decide if it's worth it or not.

src/mysql_conn.erl
```diff
@@ -616,13 +618,18 @@ named_prepare(Name, Query, State) ->
...
+handle_query_call_result([], _Type, _N, _Query, State, []) ->
     {ok, State};
-handle_query_call_result([], _Query, State, [{ColumnNames, Rows}]) ->
+handle_query_call_result([], _Type, _N, _Query, State, [{ColumnNames, Rows}]) ->
     {{ok, ColumnNames, Rows}, State};
-handle_query_call_result([], _Query, State, ResultSetsAcc) ->
+handle_query_call_result([], _Type, _N, _Query, State, ResultSetsAcc) ->
     {{ok, lists:reverse(ResultSetsAcc)}, State};
-handle_query_call_result([Rec|Recs], Query, State = #state{transaction_levels = L}, ResultSetsAcc) ->
+handle_query_call_result([Rec|Recs], Type, N, Query, State = #state{transaction_levels = L}, ResultSetsAcc) ->
```
zuiderkwast: Maybe `Type` could have a more descriptive name, e.g. `MultiResult`. A type makes me think of things like integer, boolean and such things (i.e. datatypes).

Also line break to 80, preferably.

src/mysql_conn.erl
```diff
@@ -632,12 +639,14 @@ handle_query_call_result([Rec|Recs], Query, State = #state{transaction_levels =
             Reply = {implicit_commit, Length, Query},
             [] = demonitor_processes(L, Length),
             {Reply, State#state{transaction_levels = []}};
-        #ok{} ->
-            handle_query_call_result(Recs, Query, State, ResultSetsAcc);
-        #resultset{cols = ColDefs, rows = Rows} ->
+        #ok{status = Status} ->
+            maybe_log_slow_query(State, Status, Type, N, Query),
```
zuiderkwast: Good catch!

src/mysql_conn.erl
```diff
@@ -667,6 +676,27 @@ log_warnings(#state{socket = Socket, sockmod = SockMod}, Query) ->
...
+    IndexHint = if
+        S band ?SERVER_STATUS_NO_GOOD_INDEX_USED /= 0 ->
+            " (with no good index)";
+        S band ?SERVER_STATUS_NO_INDEX_USED /= 0 ->
+            " (with no index)";
+        true ->
+            ""
+    end,
+    QueryNumHint = if
+        Type =:= multi_result ->
```
zuiderkwast: A typical case for case.
```Erlang
QueryNumHint =
    case Type of
        multi_result ->
            ...;
        single_result ->
            ...
    end,
```

src/mysql_conn.erl
juhlig: Ok, did as you suggested, no biggie :) I had that in mind as well, but the requirement for an extra `case` was why I ended up using an extra parameter instead. So, is that what you would like it to be like?

src/mysql_conn.erl
juhlig: `Type` was removed by implementing above suggestion. Reduced line length to <80 as well.

src/mysql_conn.erl
juhlig: Replaced with a case, though it's a little different now because of the altogether removal of `Type`.

src/mysql_conn.erl
juhlig: `Type` -> obsoleted by implementing above suggestion.

Line wdth =<80 -> done

src/mysql_conn.erl
zuiderkwast: Yes, I slightly prefer the combined parameter. Nice!

zuiderkwast:
Thanks!

PR #152, opened 2019-12-19
Implement caching_sha2_password authentication

juhlig:
This implements the `caching_sha2_password` authentication mechanism, which became the default in MySQL8.

Major changes and restructuring (mainly in the `mysql_protocol` module) was necessary to achieve this, and I'm not sure that all of what I did is in fact that great, though I took great care to stay as clean as possible. In doing so, I removed some comments that I might have to add back later, and there are new places that need some clarifying comments, so this is still a WIP.

Good news is, all the tests (done locally) still pass with MySQL 5.7, and with MySQL 8 only 4 unrelated ones (for transaction deadlocks) fail. Bad news is, I'm not 100% sure that I didn't miss any corner cases, as the protocol for this is _really_ complicated, and I had to rummage in community drivers for other languages (Java mostly) to make this work.

One thing I did _not_ implement is client-side caching of the server's public key. It is not required to make things work, and I'm not sure what would really be the best way to do this. Anyway, there is nothing in the way of an implementation at a later time.

Speaking of caching, there is one thing we might want to consider when it comes to testing: only the first login of a given user performs a full authentication (until the server is either restarted or `FLUSH PRIVILEGES` is used), after that the server caches it, so all subsequent logins will take the fast path. Don't know if it is of any consequence, though.


juhlig:
In the last commit, I did some code cleanup. There is still `change_user_finish_or_switch_auth`, which is almost identical to `handshake_finish_or_switch_auth`, I'm still uncertain how best to clean that up.

I also (re-)commented some aspects of the code.

Finally, I disabled the transaction deadlock tests for MySQL 8.x.y the same way as it is already done for MySQL 5.7.x. I investigated a little there, and it looks like the rollback behavior of InnoDB in case of deadlocks has changed. The docs (including those for MySQL 5.7 and 8.0) say that the statements of the transaction are rolled back _except_ the `BEGIN`, and that further statements are executed inside the transaction. However, my investigation seems to show that this is no longer the case, the OK packet received from a statement after a deadlock has the `SERVER_STATUS_IN_TRANS` flag unset and the statement has been executed.

juhlig:
Last commit eliminates the almost-duplicate functions `handshake_finish_or_switch_auth` / `change_user_finish_or_switch_auth` by factoring out the main logic into a new function `auth_finish_or_switch` and putting handshake-specific things into a stripped-down version of `handshake_finish_or_switch_auth`.

juhlig:
In the last commit (which will probably be the last this year ;)), I changed the travis configuration to include MySQL8. It seems to fail for some older OTP releases (notably, however, it succeeds for 21 but fails for 21.1). I'll have to investigate this next year. Until then, you may play with it, happy holidays :)

src/mysql_protocol.erl
```diff
@@ -233,26 +270,27 @@ fetch_execute_response(SockModule, Socket, FilterMap, Timeout) ->
     fetch_response(SockModule, Socket, Timeout, binary, FilterMap, []).
 
 %% @doc Changes the user of the connection.
--spec change_user(module(), term(), iodata(), iodata(), binary(),
+-spec change_user(module(), term(), iodata(), iodata(), binary(), binary(),
```
zuiderkwast: Is `AuthPluginName` mandatory for change user or does an empty binary here mean that the default auth method will be used? Comments and/or edoc is welcome (also during development, it's a good habit).

The spec could be easier to read annotated with parameter names. :)

src/mysql_protocol.erl
```diff
@@ -1220,6 +1257,33 @@ parse_auth_method_switch(AMSData) ->
...
+    %% byte signals Fast Auth Success.
+    #auth_more_data{message = fast_auth_success};
+parse_auth_more_data(?authmethod_caching_sha2_password, <<4>>) ->
+    %% With caching_sha2_password authentication, a single 0x04
+    %% byte signals a Full Auth Request.
+    #auth_more_data{message = full_auth_request};
+parse_auth_more_data(?authmethod_caching_sha2_password, Data) ->
+    %% With caching_sha2_password authentication, anything
+    %% other than the above should be the public key of the
+    %% server.
```
zuiderkwast: These are good comments!

src/mysql_protocol.erl
```diff
@@ -1220,6 +1257,33 @@ parse_auth_method_switch(AMSData) ->
...
+    PubKey = case public_key:pem_decode(Data) of
+        [PemEntry = #'SubjectPublicKeyInfo'{}] ->
+            public_key:pem_entry_decode(PemEntry);
+        [PemEntry = #'RSAPublicKey'{}] ->
+            PemEntry
+    end,
+    #auth_more_data{message = {public_key, PubKey}};
+parse_auth_more_data(?authmethod_sha256_password, Data) ->
+    %% sha256_password authentication is superseded by
+    %% caching_sha2_password.
```
zuiderkwast: ... but we implement it anyway, because it's easy?

src/mysql_protocol.erl
```diff
@@ -1230,37 +1294,93 @@ get_null_terminated_binary(<<0, Rest/binary>>, Acc) ->
 get_null_terminated_binary(<<Ch, Rest/binary>>, Acc) ->
     get_null_terminated_binary(Rest, <<Acc/binary, Ch>>).
 
--spec hash_password(Password :: iodata(), Salt :: binary()) -> Hash :: binary().
-hash_password(Password, Salt) ->
+-spec hash_password(AuthMethod :: binary(), Password :: iodata(), Salt :: binary()) -> Hash :: binary().
+hash_password(AuthMethod, Password, Salt) when not is_binary(Password) ->
```
zuiderkwast: `when is_list(Password)` -- better check what it is rather than what it isn't. (It's legacy, I know, but there is no risk of breaking anything here where iolist_to_binary is used on it.)

src/mysql_protocol.erl
```diff
@@ -1230,37 +1294,93 @@ get_null_terminated_binary(<<0, Rest/binary>>, Acc) ->
...
-
--spec hash_non_empty_password(Password :: iodata(), Salt :: binary()) ->
-    Hash :: binary().
-hash_non_empty_password(Password, Salt) ->
-    Salt1 = case Salt of
-        <<SaltNoNul:20/binary-unit:8, 0>> -> SaltNoNul;
-        _ when size(Salt) == 20           -> Salt
-    end,
-    %% Hash as described above.
+    Salt1 = trim_salt(Salt),
```
zuiderkwast: Good refactoring

src/mysql_protocol.erl
```diff
@@ -1230,37 +1294,93 @@ get_null_terminated_binary(<<0, Rest/binary>>, Acc) ->
...
+    %% SHA256( password ) XOR SHA256( SHA256( SHA256( password ) ) <concat>
+    %%                                        "20-bytes random data from server" )
+    Salt1 = trim_salt(Salt),
+    <<Hash1Num:256>> = Hash1 = crypto:hash(sha256, Password),
+    Hash2 = crypto:hash(sha256, Hash1),
+    <<Hash3Num:256>> = crypto:hash(sha256, <<Hash2/binary, Salt1/binary>>),
+    <<(Hash1Num bxor Hash3Num):256>>;
+hash_password(?authmethod_sha256_password, Password, Salt) ->
+    %% sha256_password authentication is superseded by
+    %% caching_sha2_password.
```
zuiderkwast: Maybe more relevant to mention here: sha2_password is using the same password hashing as caching_sha2_password

src/mysql_protocol.erl
```diff
@@ -1230,37 +1294,93 @@ get_null_terminated_binary(<<0, Rest/binary>>, Acc) ->
...
+    RsaPadding = case ServerVersion < [8, 0, 5] of
+        true -> rsa_pkcs1_padding;
+        false -> rsa_pkcs1_oaep_padding
+    end,
+    public_key:encrypt_public(Password2, PubKey, [{rsa_padding, RsaPadding}]);
+encrypt_password(?authmethod_sha256_password, Password, Salt, PubKey,
+                 ServerVersion) ->
+    %% sha256_password authentication is superseded by
+    %% caching_sha2_password.
+    encrypt_password(?authmethod_caching_sha2_password, Password, Salt, PubKey,
```
zuiderkwast: Why does this function even have AuthMethod as a parameter? It is only used for these two and they are using the same algorithm. Maybe the AuthMethod parameter can be removed.

src/mysql_protocol.erl
```diff
@@ -1230,37 +1294,93 @@ get_null_terminated_binary(<<0, Rest/binary>>, Acc) ->
...
+    <<Hash3Num:256>> = crypto:hash(sha256, <<Hash2/binary, Salt1/binary>>),
+    <<(Hash1Num bxor Hash3Num):256>>;
+hash_password(?authmethod_sha256_password, Password, Salt) ->
+    %% sha256_password authentication is superseded by
+    %% caching_sha2_password.
+    hash_password(?authmethod_caching_sha2_password, Password, Salt);
+hash_password(UnknownAuthMethod, _, _) ->
+    error({auth_method, UnknownAuthMethod}).
+
+encrypt_password(AuthMethod, Password, Salt, PubKey, ServerVersion)
```
zuiderkwast: A comment/edoc about what (which auth methods) the function is used for would be good though.

src/mysql_protocol.erl
```diff
@@ -73,36 +81,65 @@ handshake(Username, Password, Database, SockModule0, SSLOpts, Socket0,
...
     end.
 
-handshake_finish_or_switch_auth(Handshake = #handshake{status = Status}, Password,
-                                SockModule, Socket, SeqNum0) ->
-    {ok, ConfirmPacket, SeqNum1} = recv_packet(SockModule, Socket, SeqNum0),
-    case parse_handshake_confirm(ConfirmPacket) of
+
+handshake_finish_or_switch_auth(Handshake, Password, SockModule, Socket, SeqNum) ->
+    #handshake{auth_plugin_name = AuthPluginName,
+	       auth_plugin_data = AuthPluginData,
```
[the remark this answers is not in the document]
juhlig: Oops =^^=

src/mysql_protocol.erl
```diff
@@ -73,36 +81,65 @@ handshake(Username, Password, Database, SockModule0, SSLOpts, Socket0,
...
+
+auth_finish_or_switch(AuthPluginName, AuthPluginData, Password,
+                      SockModule, Socket, ServerVersion, SeqNum0) ->
+    {ok, ConfirmPacket, SeqNum1} = recv_packet(SockModule, Socket, SeqNum0),
+    case parse_handshake_confirm(AuthPluginName, ConfirmPacket) of
+        #ok{} = Ok ->
+            Ok;
+        #auth_method_switch{auth_plugin_name = SwitchAuthPluginName,
+                            auth_plugin_data = SwitchAuthPluginData} ->
+            Hash = hash_password(SwitchAuthPluginName, Password, SwitchAuthPluginData),
```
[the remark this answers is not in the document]
juhlig: And believe me, after doing this, I would prefer not to know ;)

You may look at [this flow chart](https://dev.mysql.com/doc/dev/mysql-server/latest/page_caching_sha2_authentication_exchanges.html) if you like your brain twisted, or read [this more easily digestable text](http://www.dataarchitect.cloud/preparing-your-community-connector-for-mysql-8-part-2-sha256/). The entire things is very underdocumented from MySQLs side, most of it I got from the above two sources, odd bits and pieces from Google, and by looking at the Java driver. And part is all guesswork.

Anyway, this is a very involved thing. It would require a very extensive comment. Though, I don't know where this may be better placed.

src/mysql_protocol.erl
juhlig: Hm, partially ;) From the noncommittal comment on MySQLs website, I read this as "anything required for sha256_password can be done with the mechanisms included in caching_sha2_password". When experimenting, I sometimes received auth switches to sha256_password (though neither the server nor the user had it specified), so it looks like we must support it. This is not conclusive, repeating the very same at a later time didn't use sha256_password. In short, I don't know what is required to show up =^^=

src/mysql_protocol.erl
juhlig: Ok

src/mysql_protocol.erl
juhlig: Wasn't sure here. I included it preemptively in case more auth methods show up. OTOH, I can remove it for now and we can add it back when new ones show up. What do you say?

src/mysql_protocol.erl
juhlig: Maybe even better, we could normalize those things to binaries in the entry functions like `handshake`, and just assume we work with binaries in all the internal functions?

src/mysql_protocol.erl
juhlig: Yes, it is required. We need to use the servers' default auth method (like in handshake), but we have no handshake at this point to read it from.

Empty binary appears when the server does not know auth plugins, in which case we resort to mysql_native_password. This is how we handled that before this PR.

src/mysql_protocol.erl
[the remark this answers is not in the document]
zuiderkwast: I don't really want to get my brain twisted :-) but I want to keep the code maintainable. One sentence at the definition of each function can be good enough, just to make it easier to follow the code. At some other place, you added quite good comments. These links can be added in code comments, if useful. Guesswork/assumptions are good to document too, but just a line or so.

src/mysql_protocol.erl
zuiderkwast: Good then. :-)

src/mysql_protocol.erl
zuiderkwast: Ah, I see, the default auth method is stored in the state and passed here, so no change in the exported function for change user. That's good.

src/mysql_protocol.erl
zuiderkwast: Sure, for username and password and such things, if it reduces complexity.

For queries and other potential large strings, we better not. The IO functions can send nested lists to sockets by just traversing them, without allocating a binary.

src/mysql_protocol.erl
zuiderkwast: Remove it then. Preemptively adding stuff is rarily good. "Ain't gonna need it" as they say.

src/mysql_protocol.erl
juhlig: Yeah, I was thinking of things like username and password, things we need to work with internally, so we don't have to do normalization in multiple places. For others like queries that we just pass along, there is no gain.

In reply to the original request, I turned it around a little in the last commit and check with `is_binary`, since this is really what we _want_ it to be to work with.

src/mysql_protocol.erl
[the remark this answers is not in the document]
juhlig: Ok, I added a general comment for the entire function, and individual ones in the `case` branches. What do you think?

juhlig:
And finally, I found what was causing the test failures travis for OTP<22 :) In OTP 22, the `rsa_pad` option for `public_key:encrypt_public` was renamed to `rsa_padding`, but `rsa_pad` was kept for backwards compatibility. So by using `rsa_pad` instead of `rsa_padding`, it works for all releases that we support :)

src/mysql_protocol.erl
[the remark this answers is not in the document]
zuiderkwast: Awesome

src/mysql_protocol.erl
zuiderkwast: Yes, do this if you want. It can also be done later. This PR is almost ready to merge, right? Whenever you remove the "WIP"...

include/records.hrl
```diff
@@ -32,6 +32,10 @@
           auth_plugin_data :: binary()
          }).
 
+-record(auth_more_data, {
+          message :: term()
+         }).
```
zuiderkwast: This doesn't seem useful. Can it be a type instead: `-type auth_more_data :: fast_auth_completed | full_auth_requested | {public_key, _}.`?

Btw, the record definitions which are only used in one module are probably better to define there than in this include file, but that's another PR perhaps...

zuiderkwast:
> it looks like the rollback behavior of InnoDB in case of deadlocks has changed. The docs (including those for MySQL 5.7 and 8.0) say that the statements of the transaction are rolled back _except_ the `BEGIN`, and that further statements are executed inside the transaction. However, my investigation seems to show that this is no longer the case, the OK packet received from a statement after a deadlock has the `SERVER_STATUS_IN_TRANS` flag unset and the statement has been executed.

Ah! I suspected something like this but I didn't have time to investigate it fully. Do you think we should implement a workaround which involves checking the server version and executing an extra `BEGIN` or similar?

include/records.hrl
juhlig: You're right, this is way better :) Changed as suggested in last commit.

juhlig:
> Ah! I suspected something like this but I didn't have time to investigate it fully. Do you think we should implement a workaround which involves checking the server version and executing an extra `BEGIN` or similar?

It seems that we can _always_ do an extra `BEGIN` when retrying a transaction, regardless of the server version. Solutions to similar scenarios I saw, using procedures in MySQL itself, did something like `BEGIN; ROLLBACK` there. I have to investigate in more detail, but it seems that a `BEGIN` in an active transaction simply has no effect (not even a warning).

src/mysql_protocol.erl
juhlig: I think it is ready now, and so I remove the WIP :)

zuiderkwast:
OK, really great work. Thanks! Updating the README remains...

Issue #99, opened 2018-12-02
Test multiple MySQL and MariaDB versions


zuiderkwast:
Use Travis matrix. See e.g. https://github.com/xerions/mariaex/blob/master/.travis.yml or https://github.com/mysqljs/mysql/blob/master/.travis.yml

michaellenaghan:
Another option—as in "in addition to" rather than "in place of"—would be to use Docker the way MyXQL does [here](https://github.com/elixir-ecto/myxql/blob/master/ci.sh).

Notice that with their Docker approach you can select the version you want to test by using an environment variable, e.g.:

```
DB=mysql:5.5 ./ci.sh ...
DB=mysql:5.6 ./ci.sh ...
DB=mysql:5.7 ./ci.sh ...
```

zuiderkwast:
Nice! For Travis we can definitely go for this!

As long as it's still possible to run the tests locally without Docker.

michaellenaghan:
They just use Docker to run MySQL; tests run locally, but against a Dockerized MySQL. Very simple idea, very simple implementation.

Running against both Dockerized and not-Dockerized MySQL would be trivially easy. Off the top of my head I can think of two issues. First, you'd want to parameterize some connection info in your tests; for example, I don't think you'd want Dockerized MySQL running on port 3306. Second, you'd want to sometimes automatically create the MySQL users you currently ask people to create themselves.

zuiderkwast:
Yes I know. I just want it to be possible to run the tests using a locally installed MySQL (without docker) as well.

> I don't think you'd want Dockerized MySQL running on port 3306

Why not?

> Second, you'd want to sometimes automatically create the MySQL users you currently ask people to create themselves.

Sure, feel free to change this!

We have one test case for connecting using unix socket. That only works if the database is running locally. I don't know if this works with docker.

michaellenaghan:
>> I don't think you'd want Dockerized MySQL running on port 3306

> Why not?

I run MySQL locally. I wouldn't want to have to stop it just so that I could run mysql-otp tests. Not only that, but by making the port configurable you could in theory test different MySQL versions in parallel.

zuiderkwast:
Fixed in #157 

Future work:
* Add MariaDB 10.4 and 10.5
* Generate combined code coverage from the different test runs

Issue #85, opened 2018-06-30
OTP 20.3 TLS/SSL problem


zuiderkwast:
There seems to be a problem with OTP 20.3, not present in previous OTP 20 versions.

When this is solved, add 20.3 to the travis ci matrix.

```
=INFO REPORT==== 29-Jun-2018::05:09:20 ===
TLS client: In state certify at ssl_handshake.erl:1295 generated CLIENT ALERT: Fatal - Handshake Failure - {bad_cert,hostname_check_failed}
ssl_tests: successful_ssl_connect_test (module 'ssl_tests')...*skipped*
undefined
*unexpected termination of test process*
::{failed_to_upgrade_socket,{tls_alert,"handshake failure"}}
```

getong:
I think it is fixed in #84 , but there is still logger problem in erlang 21. I will fix it in a few days.

getong:
It can be closed now.

DenysGonchar:
this issue is still valid and it's reproducible only on OTP 20.3. please see [travis report](https://travis-ci.org/DenysGonchar/mysql-otp/jobs/542374424)
```
ssl_tests: successful_ssl_connect_test (module 'ssl_tests')...
=INFO REPORT==== 6-Jun-2019::17:08:19 ===
TLS client: In state certify at tls_connection.erl:739 generated CLIENT ALERT: Fatal - Handshake Failure - malformed_handshake_data
*skipped*
undefined
*unexpected termination of test process*
::{failed_to_upgrade_socket,{tls_alert,"handshake failure"}}
```

zuiderkwast:
OK, obviously there's a TLS error. I think this is related to some bug in OTP 20.3. Perhaps it's works in OTP 20.3 with different ssl options? I don't know if it's possible to solve here, but we can at least document it. It would be good to link this one to an OTP bug report or similar.

PR #153, opened 2019-12-27
Travis tests for MySQL 5.6 - 8.0 and MariaDB 10.3

juhlig:
#99

This is the best I could come up with. It defines an explicit job for each Ubuntu distro (thus, MySQL version, precise for 5.5, trusty for 5.6, xenial for 5.7, and xenial upgrading from 5.7 to 8.0) and OTP version (excluding >=21.0 in precise, as it seems it is not supported there).

The same can probably be done for MariaDB.

This PR is based on the caching_sha2_password PR, so I can run the tests against MySQL 8.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,5 @@
-dist: xenial
-sudo: true
+os: linux
 language: erlang
-services:
-  - mysql
-env:
-  - MYSQL8=0
-  - MYSQL8=1
 before_script:
   - if [ $MYSQL8 = 1 ]; then
       wget https://repo.mysql.com/mysql-apt-config_0.8.14-1_all.deb;
@@ -23,20 +17,58 @@ before_script:
   - cat test/ssl/my-ssl.cnf | sudo tee -a /etc/mysql/conf.d/my-ssl.cnf
   - sudo service mysql start
   - sleep 5
-  - mysql -uroot -e "CREATE USER otptest@localhost IDENTIFIED BY 'otptest';"
-  - mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptest@localhost;"
-  - mysql -uroot -e "CREATE USER otptest2@localhost IDENTIFIED BY 'otptest2';"
-  - mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptest2@localhost;"
-  - mysql -uroot -e "CREATE USER otptestssl@localhost IDENTIFIED BY 'otptestssl' REQUIRE SSL;"
-  - mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptestssl@localhost;"
+  - sudo mysql -uroot -e "CREATE USER otptest@localhost IDENTIFIED BY 'otptest';"
+  - sudo mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptest@localhost;"
+  - sudo mysql -uroot -e "CREATE USER otptest2@localhost IDENTIFIED BY 'otptest2';"
+  - sudo mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptest2@localhost;"
+  # REQUIRE SSL must be specified only in CREATE USER in MySQL >= 8.0, only in GRANT in MySQL < 5.7;
+  # MySQL 5.7 allows both variants.
+  - (sudo mysql -uroot -e "CREATE USER otptestssl@localhost IDENTIFIED BY 'otptestssl' REQUIRE SSL;" &&
+    sudo mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptestssl@localhost;") ||
+    (sudo mysql -uroot -e "CREATE USER otptestssl@localhost IDENTIFIED BY 'otptestssl';" &&
+    sudo mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptestssl@localhost REQUIRE SSL;")
 script: 'make tests'
-otp_release:
-  - 22.0
-  - 21.1
-  - 21.0
-  - 20.0
-  - 19.0
-  - 18.3
-  - 17.3
-  - 17.0
-  - R16B03-1
+jobs:
+  include:
+    - name: OTP/22.1 + MySQL8.0
+      dist: xenial
+      env: MYSQL8=1
+      services:
+        - mysql
+      otp_release: 22.1
+    - name: OTP/21.3 + MySQL5.7
+      dist: xenial
+      env: MYSQL8=0
+      services:
+        - mysql
+      otp_release: 21.3
+    - name: OTP/20.1 + MySQL5.6
+      dist: trusty
+      env: MYSQL8=0
+      services:
+        - mysql
+      otp_release: 20.1
+    - name: OTP/19.3 + MariaDB10.3
+      dist: xenial
+      addons:
+        mariadb: '10.3'
+      env: MYSQL8=0
+      otp_release: 19.3
+    - name: OTP/18.3 + MySQL8.0
+      dist: xenial
+      env: MYSQL8=1
+      services:
+        - mysql
+      otp_release: 18.3
+    - name: OTP/17.3 + MariaDB10.3
+      dist: xenial
+      addons:
+        mariadb: '10.3'
+      env: MYSQL8=0
+      otp_release: 17.3
+    - name: OTP/R16B03-1 + MySQL5.6
+      dist: trusty
+      env: MYSQL8=0
+      services:
+        - mysql
+      otp_release: R16B03-1
--- a/README.md
+++ b/README.md
@@ -124,8 +124,13 @@ GRANT ALL PRIVILEGES ON otptest.* TO otptest@localhost;
 CREATE USER otptest2@localhost IDENTIFIED BY 'otptest2';
 GRANT ALL PRIVILEGES ON otptest.* TO otptest2@localhost;
 
+# in MySQL < 5.7, REQUIRE SSL must be given in GRANT
 CREATE USER otptestssl@localhost IDENTIFIED BY 'otptestssl';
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -14,17 +23,157 @@ before_script:
   - mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptest@localhost;"
   - mysql -uroot -e "CREATE USER otptest2@localhost IDENTIFIED BY 'otptest2';"
   - mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptest2@localhost;"
-  - mysql -uroot -e "CREATE USER otptestssl@localhost IDENTIFIED BY 'otptestssl';"
-  - mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptestssl@localhost REQUIRE SSL;"
+  - (mysql -uroot -e "CREATE USER otptestssl@localhost IDENTIFIED BY 'otptestssl' REQUIRE SSL;" &&
+    mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptestssl@localhost;") ||
+    (mysql -uroot -e "CREATE USER otptestssl@localhost IDENTIFIED BY 'otptestssl';" &&
+    mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptestssl@localhost REQUIRE SSL;")
```
zuiderkwast: Wow. May I assume this the only way to make this work in all versions of MySQL? (A code comment explaining that wouldn't hurt.)

.travis.yml
```diff
@@ -14,17 +23,157 @@ before_script:
...
-  - 18.3
-  - 17.3
-  - 17.0
-  - R16B03-1
-
+jobs:
+  include:
+# OTP/22.0 probably not supported on precise
+#    - name: OTP/22.0+MySQL5.5
+#      dist: precise
```
zuiderkwast: It's totally fine to skip these combinations. I think picking one MySQL version for each OTP release would be sufficient, to keep the number of test runs to about the same as we have already.

zuiderkwast:
I forgot to say: Awesome! This is really promising!

.travis.yml
juhlig: Probably not the _only_ way but the easiest/quickest I came up with, as my main goal was to get all MySQL versions running, sort of as a proof of concept, so I left this part in a temporary mess ;)

juhlig:
> Is there any chance we can test MySQL 8 without `caching_sha2_password` (e.g. be explicitly configuring users to use mysql_native_password) and disable the caching_sha2_password tests on servers where it's not supported? Or is it easier/better to use the unchanged default authentication for each version?

Hm, yes, for that the default auth method needs to be changed/overridden to mysql_native in a config file (can't be done with a variable, its read-only), and the test users must be changed to use mysql_native as their auth method, too.

Roughly, neither way seems more complicated than the respective other, I'd say :)

.travis.yml
juhlig: Oh, btw, I wanted to ask why those OTP releases in particular were picked for testing?

juhlig:
> I forgot to say: Awesome! This is really promising!

Thanks :)

.travis.yml
zuiderkwast: No particular reason. I thought NN.0 would be good in general, but maybe the latest minor version per major version would make more sense.

Another option would be to pick the versions included of various (old) releases of Debian and other distros.

We can change them. Do you have a suggestion?

juhlig: Hm, I suggest we go for the latest minor versions (usually, the *.3s for past releases, and the respective latest minor for the latest major release), where most of the bugs of that major version have supposedly been fixed, so I'd expect developers to converge on them over time. If there are any particular releases containing a bug that affected us and that we had to work around, we may want to include those as well.

juhlig:
> > Is there any chance we can test MySQL 8 without `caching_sha2_password` (e.g. be explicitly configuring users to use mysql_native_password) and disable the caching_sha2_password tests on servers where it's not supported? Or is it easier/better to use the unchanged default authentication for each version?
> 
> Hm, yes, for that the default auth method needs to be changed/overridden to mysql_native in a config file (can't be done with a variable, its read-only), and the test users must be changed to use mysql_native as their auth method, too.
> 
> Roughly, neither way seems more complicated than the respective other, I'd say :)

On second thought... what are you aiming at with your question? :) Run all tests with MySQL 8, or additional tests for (some of) server/user auth combinations set to sha2/sha2, sha2/native, native/sha2, native/native?

.travis.yml
zuiderkwast: OK! Only keep #85 in mind then.

zuiderkwast:
> 
> 
> > > Is there any chance we can test MySQL 8 without `caching_sha2_password` (e.g. be explicitly configuring users to use mysql_native_password) and disable the caching_sha2_password tests on servers where it's not supported? Or is it easier/better to use the unchanged default authentication for each version?
> > 
> > 
> > Hm, yes, for that the default auth method needs to be changed/overridden to mysql_native in a config file (can't be done with a variable, its read-only), and the test users must be changed to use mysql_native as their auth method, too.
> > Roughly, neither way seems more complicated than the respective other, I'd say :)
> 
> On second thought... what are you aiming at with your question? :) Run all tests with MySQL 8, or additional tests for (some of) server/user auth combinations set to sha2/sha2, sha2/native, native/sha2, native/native?

I was mostly thinking about testing MySQL 8 before caching_sha2 was implemented and merged, which is a non-issue now. :-)

Other than that, I only tried to keep the different factors independent of other factors and e.g. test one auth method in one test case and another auth method in another test case.

It is more important to run the full regression using the default auth method on each MySQL version, so please disregard this. (We can keep it in mind as an option for testing future non-default auth methods though, perhaps.)

juhlig:
Ok, here are some things I encountered when trying for MariaDB (in a separate branch, to avoid triggering travis over and over, will tidy it up on monday). In a nutshell, I now have a working travis setup for OTP versions from 16B03-1 through all *.3 (exception see below) to 22.1 and MySQL versions from 5.6 to 8.0, plus MariaDB 10.3 with OTP/22.1.

* travis is currently decomissioning precise, so I removed the tests using that distro, so MySQL 5.5 is out
* there seems to be no (easy) way to fix or work around the TLS bug in OTP/20.3, and OTP/20.2 is not available on travis for xenial, so instead of 20.3, I set up 20.1
* with MariaDB 10.4, the SSL tests do not work (as mentioned in #157) for some obscure reason, probably something to do with TLSv1.3. For that reason, I set up the test to use 10.3 instead where it still works.

In total, that amounts to 22 tests, and we may want to reduce that number as you suggested above. What do you think we should pair up and keep?

zuiderkwast:
Awesome!

Let's merge what you have working now (after tidying up) and leave the rest for the future. It's good to document the problems by updating/creating relevant issues.

As for the {OTP, DB} combinations, how about this?

| OTP | DB |
|-----|----|
| 22.1 | My 8 |
| 21.3 | My 5.7 |
| 20.1 | My 5.6 |
| 19.3 | Maria 10.3 |
| 18.3 | My 8 |
| 17.3 | Maria 10.3 |
| R16B03 | My 5.6 |

Isn't it enough?

juhlig:
Ok, did as suggested :) Tests all pass, too. Also put a comment over the `... REQUIRE SSL...` part to explain. If you wonder about the `sudo` I before the `mysql -e ...`, that seems to be required from MariaDB 10.4 onwards. Locally, I had to do the same for MySQL 8, but in travis that was not necessary.

juhlig:
I also updated the README part for the test setup, concerning `REQUIRE SSL` in `GRANT` (MySQL < 5.7) vs `CREATE USER` (MySQL >= 8.0)

README.md
```diff
@@ -124,8 +124,13 @@ GRANT ALL PRIVILEGES ON otptest.* TO otptest@localhost;
 CREATE USER otptest2@localhost IDENTIFIED BY 'otptest2';
 GRANT ALL PRIVILEGES ON otptest.* TO otptest2@localhost;
 
+# in MySQL < 5.7, REQUIRE SSL must be given in GRANT
```
zuiderkwast: Good comments.

Is this MySQL-specific comment syntax? In standard SQL, the comments start with `--`. I've only used the standard one; never bothered to learn the non-standard one(s).

.travis.yml
```diff
@@ -23,20 +17,58 @@ before_script:
...
-  - mysql -uroot -e "CREATE USER otptest2@localhost IDENTIFIED BY 'otptest2';"
-  - mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptest2@localhost;"
-  - mysql -uroot -e "CREATE USER otptestssl@localhost IDENTIFIED BY 'otptestssl' REQUIRE SSL;"
-  - mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptestssl@localhost;"
+  - sudo mysql -uroot -e "CREATE USER otptest@localhost IDENTIFIED BY 'otptest';"
+  - sudo mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptest@localhost;"
+  - sudo mysql -uroot -e "CREATE USER otptest2@localhost IDENTIFIED BY 'otptest2';"
+  - sudo mysql -uroot -e "GRANT ALL PRIVILEGES ON otptest.* TO otptest2@localhost;"
+  # REQUIRE SSL must be specified only in CREATE USER in MySQL >= 8.0, only in GRANT in MySQL < 5.7;
+  # MySQL 5.7 allows both variants.
```
zuiderkwast: Good comments!

zuiderkwast:
Awesome!

README.md
juhlig: Uh, looks like it is: https://dev.mysql.com/doc/refman/8.0/en/comments.html

I for my part wasn't aware of `#` being MySQL-specific, so I used it without thinking.

juhlig: Opened another PR #159 to use the Standard SQL style.

zuiderkwast: Then you've learned something. :+1:

juhlig: True :)

PR #156, opened 2019-12-28
Workaround for changed deadlock handling in MySQL 5.7+

zuiderkwast:
According to the MySQL docs, when a deadlock error occurs, the
transaction is rollbacked but the BEGIN statement is not cancelled.
This is the case up to MySQL 5.6 but in 5.7, the BEGIN statement
needs to be executed again when restarting a transaction. (The
official MySQL documentation has not been updated to reflect this.)


juhlig:
FYI: https://bugs.mysql.com/bug.php?id=98133

zuiderkwast:
Thanks Mr. Uhlig! :-)

PR #157, opened 2019-12-29
Make tests pass with MariaDB 10.3.15

zuiderkwast:
MariaDB sometimes returns error codes different to those
returned by MySQL. This commit adjusts a few test cases so
that they accept both error codes, where they differ.

### DIFF
--- a/test/mysql_tests.erl
+++ b/test/mysql_tests.erl
@@ -1006,8 +1006,7 @@ is_access_denied({1045, <<"28000">>, <<"Access denie", _/binary>>}) ->
 is_access_denied({1698, <<"28000">>, <<"Access denie", _/binary>>}) ->
     true; % MariaDB 10.3.15
 is_access_denied({1251, <<"08004">>, <<"Client does not support authentication "
-                                       "protocol requested by server; consider "
-                                       "upgrading MariaDB client">>}) ->
+                                       "protocol requested", _/binary>>}) ->
     true; % This has been observed with MariaDB 10.3.13
 is_access_denied(_) ->
     false.
--- a/test/mysql_tests.erl
+++ b/test/mysql_tests.erl
@@ -66,7 +66,9 @@ connect_asynchronous_failing_test() ->
             {ok, Pid} = mysql:start_link([{user, "dummy"}, {password, "junk"},
                                           {connect_mode, asynchronous}]),
             receive
-                {'EXIT', Pid, {error, {1045, <<"28000">>, _}}} -> ok
+                {'EXIT', Pid, {error, Error}} ->
+                    true = is_access_denied(Error),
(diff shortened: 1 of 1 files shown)


test/mysql_tests.erl
```diff
@@ -1005,3 +1000,14 @@ parse_db_version(Version) ->
...
   lists:map(fun binary_to_integer/1,
             binary:split(Version1, <<".">>, [global])).
+
+is_access_denied({1045, <<"28000">>, <<"Access denie", _/binary>>}) ->
+    true; % MySQL 5.x, etc.
+is_access_denied({1698, <<"28000">>, <<"Access denie", _/binary>>}) ->
+    true; % MariaDB 10.3.15
+is_access_denied({1251, <<"08004">>, <<"Client does not support authentication "
+                                       "protocol requested by server; consider "
+                                       "upgrading MariaDB client">>}) ->
```
juhlig: Without re-checking (still on my mobile instead of a real computer), the same error message (with "MySQL" instead of "MariaDB") occurs if MySQL 8 is used and the default auth plugin is set to `mysql_native_password` (vs `caching_sha2_password`) on the server. While this is probably of no consequence for current tests, it may be worth to truncate the expected message before "MariaDB", or at least to keep it in mind for the future.

zuiderkwast: Good point! I think it's good to make sure the tests pass with this configuration even though we don't run it with travis.

zuiderkwast: Good point! I think it's good to make sure the tests pass with this configuration even though we don't run it with travis.

juhlig:
I'm playing with MariaDB and travis ATM, and I can confirm that all the tests pass with these changes when MariaDB 10.3 is used. Since this is the goal of this PR, all is good :)

With MariaDB 10.4, however, the SSL test fails:
```erlang
::{failed_to_upgrade_socket,
      {tls_alert,
          {bad_record_mac,
              "TLS client: In state hello received UNDEFINED ALERT: Fatal - Bad Record MAC\n"}}}
```
I have no _real_ idea why, this is notoriously hard to debug, and I'm not exactly at home with all the encryption acrobatics ;) But strange thing is, if you include `versions` in the ssl options there, it works, even if you re-iterate everything returned in the `available` tuple from `ssl:version/0`, or what is returned in the `supported` tuple, which is supposedly what is used by default if you omit the `versions` option, at least as I read it...

zuiderkwast:
OK, I did these changes just because I happen to have MariaDB 10.3.15 installed locally.

That TLS error is something I would avoid touching unless we know what and why. ^^

I'm happy if we can get travis running mysql 5.5, 5.6, 5.7, 8 first. Then we can try to include mariadb later.

PR #159, opened 2020-01-06
Use Standard-SQL style comments

juhlig:
Replace the MySQL-specific `#` style comments with Standard-SQL `--` style comments.


zuiderkwast:
Haha, good, but it's not that important. :-) Btw, there are some other minor fixes I want to do in the README and then tag 1.6.0. I'll do that in one commit, I think.

zuiderkwast:
Solved in another commit on master

PR #163, opened 2020-05-29
Allow activation of experimental socket backend for inet/gen_tcp (OTP/23)

juhlig:
OTP/23 comes with an experimental `socket` backend for `inet`/`gen_tcp` (https://www.erlang.org/news/140), which can be activated by passing `{inet_backend, socket}` _as the first option_ in the connect options. This is currently not possible with `mysql-otp` because the `sanitize_tcp_opts` function prepends some own options to that list: https://github.com/mysql-otp/mysql-otp/blob/1a231bf672f08deae0c82420bda8bb214726d572/src/mysql_conn.erl#L171-L187 

The changes in this PR remove the `inet_backend` option from the front of the given options if present, and put it back in place after prepending the other options. This way, users can experiment with the `socket` backend in `mysql-otp`.

### DIFF
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -168,6 +168,11 @@ connect_socket(#state{tcp_opts = TcpOpts, host = Host, port = Port} = State) ->
 
     {ok, State#state{socket = Socket}}.
 
+sanitize_tcp_opts([{inet_backend, _} = InetBackend | TcpOpts0]) ->
+    %% This option is be used to turn on the experimental socket backend for
+    %% gen_tcp/inet (OTP/23). If given, it must remain the first option in the
+    %% list.
+    [InetBackend | sanitize_tcp_opts(TcpOpts0)];
 sanitize_tcp_opts(TcpOpts0) ->
     TcpOpts1 = lists:filter(
         fun


src/mysql_conn.erl
```diff
@@ -169,7 +169,13 @@ connect_socket(#state{tcp_opts = TcpOpts, host = Host, port = Port} = State) ->
     {ok, State#state{socket = Socket}}.
 
 sanitize_tcp_opts(TcpOpts0) ->
-    TcpOpts1 = lists:filter(
+    {FirstOpts, TcpOpts1} = case TcpOpts0 of
```
zuiderkwast, line 172: Please add a comment about why this is kept as the first option.

This function is becoming a bit complex. Can it be simplified in some way?

One idea:

```Erlang
sanitize_tcp_opts([{inet_backend, _} = InetBackend | TcpOpts0]) ->
    %% This option must remain the first option in the list
    [InetBackend | sanitize_tcp_opts(TcpOpts0)];
sanitize_tcp_opts(TcpOpts0) ->
   %% ...
```

juhlig:
> Btw, is it worth documenting which options are allowed, which are ignored and which have a default value different to that of gen_tcp? It's not the scope of this PR but it's perhaps worth considering, the more complexity we add here.

Hm, maybe, but in that case I think we just need to highlight the _differences_ from `gen_tcp` (ignored/different defaults), not what is the same (ie, allowed).

src/mysql_conn.erl
juhlig: Oh, good idea, I was indeed struggling a bit with that ;) Nice.

As for the explaining comment, I'll see to it.

juhlig: Done :)

juhlig:
> Before submitting, I suggest renaming this to "Allow activation of ...". Otherwise, it looks like we're enabling it by default.

Makes sense, done :)

zuiderkwast:
Feel free to merge!

PR #165, opened 2020-06-08
Improve typing

juhlig:
I just tried to write a pool adapter for mysql-otp and hnc, like mysql-otp-poolboy, for fun mostly ;)

While the actual implementation went surprisingly easy, speccing the functions turned out to be cumbersome because some types weren't exported, some even local to the functions, etc, so I'd have to duplicate them.

This is a first attempt (basis for discussion) for improving the typing in mysql-otp. I went all out, some of it can probably be dropped, some changed or improved further. Let me know what you think.

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -35,7 +35,9 @@
          transaction/2, transaction/3, transaction/4,
          change_user/3, change_user/4, reset_connection/1]).
 
--export_type([connection/0, server_reason/0, query_result/0]).
+-export_type([option/0, connection/0, query/0, statement_name/0,
+              statement_ref/0, query_param/0, query_filtermap_fun/0,
+              query_result/0, transaction_result/1, server_reason/0]).
 
 %% A connection is a ServerRef as in gen_server:call/2,3.
 -type connection() :: Name :: atom() |
@@ -48,19 +50,48 @@
 -type server_reason() :: {Code :: integer(), SQLState :: binary() | undefined,
                           Message :: binary()}.
 
--type column_names() :: [binary()].
+-type column_name() :: binary().
+-type query() :: iodata().
 -type row() :: [term()].
--type rows() :: [row()].
+
+-type query_param() :: term().
 
 -type query_filtermap_fun() :: fun((row()) -> query_filtermap_res())
-                             | fun((column_names(), row()) -> query_filtermap_res()).
+                             | fun(([column_name()], row()) -> query_filtermap_res()).
 -type query_filtermap_res() :: boolean()
                              | {true, term()}.
 
+-type statement_id() :: integer().
+-type statement_name() :: atom().
+-type statement_ref() :: statement_id() | statement_name().
+
 -type query_result() :: ok
-                      | {ok, column_names(), rows()}
-                      | {ok, [{column_names(), rows()}, ...]}
+                      | {ok, [column_name()], [row()]}
+                      | {ok, [{[column_name()], [row()]}, ...]}
                       | {error, server_reason()}.
+
+-type transaction_result(Result) :: {atomic, Result} | {aborted, Reason :: term()}.
+
+-type server_name() :: {local, Name :: atom()}
+                     | {global, GlobalName :: term()}
+                     | {via, Via :: module(), ViaName :: term()}.
+
+-type option() :: {name, ServerName :: server_name()}
+                | {host, inet:socket_address() | inet:hostname()} | {port, integer()}
+                | {user, iodata()} | {password, iodata()}
+                | {database, iodata()}
+                | {connect_mode, synchronous | asynchronous | lazy}
+                | {connect_timeout, timeout()}
+                | {log_warnings, boolean()}
+                | {log_slow_queries, boolean()}
+                | {keep_alive, boolean() | timeout()}
+                | {prepare, [{StatementName :: statement_name(), Statement :: query()}]}
+                | {queries, [query()]}
+                | {query_timeout, timeout()}
+                | {found_rows, boolean()}
+                | {query_cache_time, non_neg_integer()}
+                | {tcp_options, [gen_tcp:connect_option()]}
+                | {ssl, term()}.
 
 -include("exception.hrl").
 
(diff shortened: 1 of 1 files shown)


juhlig:
> I get the need for exporting the types, but at the same time, moving the types from the spec (e.g. the spec for start_link including the options directly) to a type which refers to another type (`options() :: [option()]`), adds indirection which means complexity.

Yeah, for the sake of discussion, I mostly did what is possible, which often is more than is needed (or feasible).

> (Btw, the EDoc for the options will still remain at `start_link`, which is good, right?)

I would leave it there, yes. I'm not sure it is possible to document individual types or parts of them. I have seen it done (`ssl` seems to do it a lot), but I would have to look up how they really did it.

> I'd like to balance these needs, not adding more steps of indirection than necessary. For example, we can skip the plural forms of the types to reduce the indirection by one step. Can you try to reduce it a little?

Sure :)

src/mysql.erl
```diff
@@ -48,20 +50,53 @@
 -type server_reason() :: {Code :: integer(), SQLState :: binary() | undefined,
                           Message :: binary()}.
 
--type column_names() :: [binary()].
+-type column_name() :: binary().
+-type column_names() :: [column_name()].
```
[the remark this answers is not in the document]
juhlig: Sure :)

src/mysql.erl
```diff
@@ -48,20 +50,53 @@
...
+-type query() :: iodata().
 -type row() :: [term()].
 -type rows() :: [row()].
 
 -type query_filtermap_fun() :: fun((row()) -> query_filtermap_res())
                              | fun((column_names(), row()) -> query_filtermap_res()).
 -type query_filtermap_res() :: boolean()
                              | {true, term()}.
+-type query_filtermap() :: no_filtermap_fun | query_filtermap_fun().
```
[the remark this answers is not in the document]
juhlig: Sounds good.

I understand that as meaning we move everything from `query/5` into the new helper function, which accepts the `no_*` atoms in the respective places, and call that from `query/2,3,4,5`, where we drop the `no_*` values from the specs, but still allow them in the implmenetation in case someone uses them there (so we don't break anything that uses them, just generate dialyzer warnings). And similar for `execute/3,4,5`. Yes?

juhlig:
The last commit implements most of the suggested changes:

The plural forms of types were removed (`options()`, `column_names()`, `rows()`) to reduce indirection.

`query/2,3,4,5` and `execute/3,4,5` delegate to `query_helper/5` and `execute_helper/5`, respectively, and `no_params`, `no_filtermap_fun` and `default_timeout` were removed from the specs (and docs) of `query` and `execute`. This in turn allows us to get rid of the additional `query_filtermap()` and `query_timeout()` types.

@zuiderkwast What do you think?

juhlig:
> (I know removing those `no_params`, `default_timeout` and `no_filtermap_fun` from the guards in query/2,3,4 would break backward compatibility, but is anyone using those?

No idea... but they have been in the docs, at least, so users may know about them at least.

> Perhaps we can allow them in query/5 and remove them from query/2,3,4 and call query_call directly in those.........

Hm, that seems inconclusive.

> First step is to disallow them by the specs I guess.)

Yeah, for now we don't break anything that way, just nag ;)

src/mysql.erl
```diff
@@ -391,18 +397,28 @@ query(Conn, Query, Params, FilterMap) when (Params == no_params orelse
...
+    when Conn :: connection(),
+         Query :: query(),
          Timeout :: default_timeout | timeout(),
-         Params :: no_params | [term()],
+         Params :: no_params | [query_param()],
          FilterMap :: no_filtermap_fun | query_filtermap_fun(),
          Result :: query_result().
-query(Conn, Query, no_params, FilterMap, Timeout) ->
+query_helper(Conn, Query, no_params, FilterMap, Timeout) ->
     query_call(Conn, {query, Query, FilterMap, Timeout});
```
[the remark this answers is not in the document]
juhlig: Hm, nothing clever comes to mind here, either =^^=

juhlig:
After a few days consideration, no improvements in terms of `no_params`,/`no_filtermap_fun`/`default_timeout` and the extra `query_helper` clause came to mind, so I guess it will have to stay this way for now.

As for the exported stuff, a friend pointed out that we may want to have another (exported) type for the result of a transaction (which is different from the result of a query), like `transaction_result() :: {atomic, term()} | {aborted, term()}`.
I'm in favor of this as it would make the (exported) typing more consistent, what do you think?

zuiderkwast:
OK, but then it should be `{atomic, result()} | {aborted, ...reason()...}` or whaterver the error may be. Does mnesia export this type?

juhlig:
No, `mnesia` doesn't export it. Even if it did, I don't think we should use it directly in `mysql-otp`. The transaction facility is modeled on `mnesia`, but really it is something different.

My last commit introduces and exports a new type `transation_result(Result) :: {atomic, Result} | {aborted, Reason :: term()}`, which is the same as `mnesia` has for it's `t_result` type.

I think this also concludes this PR, I see no more possibilities for further improvements right now. So if after a final review, you have no objections, I think this can be merged :)

src/mysql.erl
```diff
@@ -545,15 +574,20 @@ in_transaction(Conn) ->
 
 %% @doc This function executes the functional object Fun as a transaction.
 %% @see transaction/4
--spec transaction(connection(), fun()) -> {atomic, term()} | {aborted, term()}.
+-spec transaction(Conn, TransactionFun) -> TransactionResult
+    when Conn :: connection(),
+         TransactionFun :: fun((...) -> Result),
```
zuiderkwast, line 579: `fun(() -> Result)` would be more exact here, since `Args = []` when left out.

src/mysql.erl
```diff
@@ -48,20 +50,49 @@
...
+                | {log_warnings, boolean()}
+                | {log_slow_queries, boolean()}
+                | {keep_alive, boolean() | timeout()}
+                | {prepare, [{StatementName :: statement_name(), Statement :: query()}]}
+                | {queries, [query()]}
+                | {query_timeout, timeout()}
+                | {found_rows, boolean()}
+                | {query_cache_time, non_neg_integer()}
+                | {tcp_options, [gen_tcp:connect_option()]}
+                | {ssl, term()}.
```
zuiderkwast, line 94: The `term()` here can certainly be improved, e.g. `[ssl:tls_client_option()]`, but that might be a different topic.

src/mysql.erl
```diff
@@ -314,19 +324,17 @@ query(Conn, Query, Params, FilterMap) when (Params == no_params orelse
 %% statement. A prepared statement is created, executed and then cached for a
 %% certain time (specified using the option `{query_cache_time, Milliseconds}'
 %% to `start_link/1'). If the same query is executed again during this time,
-%% it does not need to be prepared again. If `Params' is omitted (or specified
-%% as `no_params'), the query is executed as a plain query. To force a query
-%% without parameters to be executed as a prepared statement, an empty list can
-%% be used for `Params'.
+%% it does not need to be prepared again. If `Params' is omitted, the query
+%% is executed as a plain query. To force a query without parameters to be
+%% executed as a prepared statement, an empty list can be used for `Params'.
```
zuiderkwast, line 329: Thanks for simplifying this!

src/mysql.erl
```diff
@@ -314,19 +324,17 @@ query(Conn, Query, Params, FilterMap) when (Params == no_params orelse
...
 %%
 %% If `FilterMap' (a fun) is specified, the function is applied to each row to
 %% filter or perform other actions on the rows, in a way similar to how
 %% `lists:filtermap/2' works, before the result is returned to the caller. See
 %% below for details.
 %%
 %% `Timeout' specifies the time to wait for a response from the database. If
-%% omitted (or specified as `default_timeout'), the timeout given in
-%% `start_link/1' is used.
+%% omitted, the timeout given in `start_link/1' is used.
```
zuiderkwast, line 337: and this! 👍

juhlig:
>Then perhaps we should consider not exporting it either. What I meant by this question.

Ah, got that wrong then.

But well, `mnesia` doesn't export _any_ types at all, so I wouldn't take that as a reference without a grain of salt ;)

> I have seen mentions in OTP (PRs or bug tracker) that you should be careful exporting too many types. That's why I'm slightly reluctant to adding many types.

I went googling for that a bit, but couldn't find anything. It was mentioned that the types should be as "tight" as possible, though.

src/mysql.erl
juhlig: While I'm in favor of this, please re-read #126. The typing in `ssl` changed between OTP versions in the 21 releases, so if we do that, it won't work everywhere.

src/mysql.erl
juhlig: Right, I'll change it.

juhlig: Done

Issue #164, opened 2020-06-06
What’s meaning of `server_name_indication`?


wwhai:
I found a SSL option named `server_name_indication` but I don’t known it’s meaning.Please publish a document for ssl configuration . Thanks.

juhlig:
The SSL options in mysql-otp are passed to [ssl:connect/3](http://erlang.org/doc/man/ssl.html#connect-3) in order to upgrade the TCP socket to SSL (the first variant of connect/3 in the documentation), they are not specific to mysql-otp. Thus, I would refer you to the [documentation for the `ssl` application in Erlang](http://erlang.org/doc/man/ssl.html).

That said, SNI is mainly of use for virtual hosting (simply, multiple virtual hosts under one IP address) in order to let a server know which virtual hosts the client wants to access _before_ the SSL connection is established (again simply, to know which is the right certificate to send). You can read more about how it works and why it is necessary in virtual hosting [here](https://en.wikipedia.org/wiki/Server_Name_Indication).

I suggest just leaving it at the default, ie not giving it in the SSL options at all. I have no idea if it has any impact for a connection to a MySQL server, since it has no virtual hosting capabilities to begin with. I guess SNI is more or less ignored, but I don't know.
Now, the documentation for `ssl:connect/2,3` says that `server_name_indication` _should_ be given when `{verify, verify_peer}` is given, so there _may_ be a problem (in which case, please give us some feedback in this ticket), which should be solved by either disabling `server_name_indication` or giving the same host as in the options given to `mysql:start_link/1`.

zuiderkwast:
Refer to the ssl documentation for Erlang/OTP: https://erlang.org/doc/man/ssl.html

In short, `server_name_indication` serves two purposes:

1. Makes the client send SNI to the server in the client hello, so that the server can choose a certificate based on this.
2. Enables host name checking when checking the certificate received from the server. `{verify, verify_peer}` makes the client check not only that the certificate is signed by a trusted CA but also that the identity of the certificate (a subjectAltName field or the CN in the Subject field in the certificate) is that of the requested service, that is, matching the value used for SNI. This is to prevent man-in-the-middle attacks.

    If `{verify, verify_peer}` is given but `server_name_indication` is not given, the server IP is required as an identity in the certificate in subjectAltName, which possible but it's not how you generally create certificates.

    To disable the identity check, pass `{server_name_indication, false}`. Note that this is not the same as leaving out the `server_name_indication` option altogether.

[...]

PR #166, opened 2020-06-11
Default SNI to hostname for SSL connections

juhlig:
In response to #164.

While `mysql-otp` sets the `verify` option for SSL connections so that it defaults to `verify_peer`, it does not set any default for `server_name_indication`. Since `ssl:connect/3` is used in the variant that upgrades a TCP socket to SSL, according to the documentation, this causes the SNI option to default to the IP address of the peer, which is probably not what is wanted.

This PR sets up a default for `server_name_indication` as well, defaulting it to the `host` parameter that may be given in the options to `mysql:start_link/1`. From an outside perspective, this has the benefit that the SSL connection process of `mysql-otp` behaves more like `ssl:connect/3` in the variant where it _open_ a SSL connection (vs upgrading an already connected TCP socket) and `server_name_indication` defaults to the given host name (instead of the peer IP address).

If the host given in the `mysql:start_link/1` options is itself an IP address (either in tuple or string form) or an atom (like `loopback`), `server_name_indication` defaults to `disable`. This is different from the `ssl:connect/3` behavior (as outlined above). While this is arguably the more intuitive way to go, it differs from what `ssl` does and would need extra documentation. So I'm putting this up for debate: if the host given in `mysql:start_link/1` is an IP address, should `server_name_indication` default to `disable` (and document that), or should no default be set then and `ssl` be allowed to process it the way it usually does (no additional documentation needed then)?

### DIFF
--- a/src/mysql_protocol.erl
+++ b/src/mysql_protocol.erl
@@ -27,7 +27,7 @@
 %% @private
 -module(mysql_protocol).
 
--export([handshake/7, change_user/8, quit/2, ping/2,
+-export([handshake/8, change_user/8, quit/2, ping/2,
          query/4, query/5, fetch_query_response/3,
          fetch_query_response/4, prepare/3, unprepare/3,
          execute/5, execute/6, fetch_execute_response/3,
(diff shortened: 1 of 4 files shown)


src/mysql_protocol.erl
```diff
@@ -406,26 +408,37 @@ server_version_to_list(ServerVersion) ->
...
 
+-spec get_default_sni(Host) -> disable | Host
+    when Host :: inet:socket_address() | inet:hostname().
+get_default_sni(Host) when is_list(Host) ->
+    case inet:parse_address(Host) of
+        {ok, _} -> disable;
+        {error, einval} -> Host
+    end;
+get_default_sni(_) ->
+    disable.
```
zuiderkwast, line 440: Also here, I'm not sure we should use `disable` (unless we match explicitly on `{local, _}` for unix sockets perhaps, but why use TLS for unix sockets?) A `gen_tcp` host can also be an atom, which I think means non-fully qualified domain name.

What I mean is that adding `disable` potentially lowers the security, which I think we shouldn't do by default if there is any doubt. It can be done by the user.

zuiderkwast:
FYI: I created the wiki page https://github.com/mysql-otp/mysql-otp/wiki/TLS which we can update once this is submitted and we have tagged a new version. We can also consider changing or removing the default `{versions, [tlsv1]}`.

src/mysql_protocol.erl
juhlig: Changed in last commit.

The atom form of the hostname, I had to dig around for that a bit, but finally found out that it can (only?) be the local hostname. I think this doesn't make much sense, and it seems to be practically undocumented, so I would guess that nobody would use that.
Other atoms that can be used (although they stem from the `ip_address` type, not `hostname`) are `loopback` and `any`, which I would treat like an IP adress. We don't have many options here, anyway, except maybe defaulting to `"localhost"` in case of `loopback`...

juhlig:
> Mentioning this in EDoc for start_link/1 would be nice.

Ok, I added a note about `verify` defaulting to `verify_peer` and `server_name_indication` defaulting to the `host` option if it is a hostname string. You ok with the phrasing?

I also added a more elaborate commit message.

zuiderkwast:
Yes, very good!

juhlig:
> Yes, very good!

This means ready for merge? ;)

zuiderkwast:
Could have been merged long ago. :) EDoc, etc. can always be added/changed later.

juhlig:
> EDoc, etc. can always be added/changed later.

Well, knowing how I tend to forget those things if I don't do them in the same PR... later --> never =^^=

PR #168, opened 2020-06-29
Support LOAD DATA LOCAL revised

juhlig:
#38

This PR adds support for LOAD DATA LOCAL, a long-standing issue. The implementation reads and sends the file in chunks, instead of reading it into memory entirely, to avoid memory problems. Sorry @stuart-thackray, I kinda stole your contribution #103 here. But I noticed that it had acquired some dust over time, and I also noticed that it had some issues to it, so I thought it best to start from scratch, with inspiration taken from your PR.

There is, however, a glitch to it, when the requested file could not be found or opened. In that case, an empty packet must be sent to the server, ie there is no distinction on the server side between an empty and a non-existing/unreadable (on the client side) file, so the error goes unnoticed. While the error may be noticed on the client side. we don't have any facility for reporting client errors to the user (yet).

Another problem popped up while I was working on this PR, but I will write a separate issue for this.

### DIFF
--- a/include/records.hrl
+++ b/include/records.hrl
@@ -39,8 +39,8 @@
              warning_count :: integer(),
              msg :: binary()}).
 %% Error packet, commonly used in the protocol.
--record(error, {source = server :: server | client, code :: integer(),
-		state :: binary() | undefined, msg :: binary()}).
+-record(error, {code :: integer(), state :: binary() | undefined,
+                msg :: binary()}).
 
 %% EOF packet, commonly used in the protocol.
 -record(eof, {status, warning_count}).
--- a/include/records.hrl
+++ b/include/records.hrl
@@ -39,8 +39,8 @@
              warning_count :: integer(),
              msg :: binary()}).
 %% Error packet, commonly used in the protocol.
--record(error, {code :: integer(), state :: binary() | undefined,
-                msg :: binary()}).
+-record(error, {source = server :: server | client, code :: integer(),
+		state :: binary() | undefined, msg :: binary()}).
 
 %% EOF packet, commonly used in the protocol.
 -record(eof, {status, warning_count}).
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -82,7 +82,7 @@
                 | {database, iodata()}
                 | {connect_mode, synchronous | asynchronous | lazy}
                 | {connect_timeout, timeout()}
-                | {local_files, disallow | allow}
+                | {allowed_local_paths, [file:name_all()]}
                 | {log_warnings, boolean()}
                 | {log_slow_queries, boolean()}
                 | {keep_alive, boolean() | timeout()}
@@ -143,14 +143,12 @@
 %%   </dd>
 %%   <dt>`{connect_timeout, Timeout}'</dt>
 %%   <dd>The maximum time to spend for start_link/1.</dd>
-%%   <dt>`{local_files, disallow | allow}'</dt>
-%%   <dd>If set to `allow', the client will advertise the `CLIENT_LOCAL_FILES'
-%%       capability in the handshake and send local data on request, for example
-%%       as part of a `LOAD DATA LOCAL INFILE' query. As this is a potential threat
-%%       that would allow hijacked servers to request any file from the client machine
-%%       to which the client has access to, the default is `disallow', so the capability
-%%       is not advertised and any such requests from the server result in an `exit'
-%%       exception with reason `unexpected_local_infile_request'.</dd>
+%%   <dt>`{allowed_local_paths, [file:name_all()]}'</dt>
+%%   <dd>This option allows you to specify a list of directories or individual
+%%       files on the client machine which the server may request, for example
+%%       when executing a `LOAD DATA LOCAL INFILE' query. Only absolute paths
+%%       are allowed. The default is an empty list, meaning the client will
+%%       not send any local files to the server.</dd>
 %%   <dt>`{log_warnings, boolean()}'</dt>
 %%   <dd>Whether to fetch warnings and log them using error_logger; default
 %%       true.</dd>
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -82,7 +82,7 @@
                 | {database, iodata()}
                 | {connect_mode, synchronous | asynchronous | lazy}
                 | {connect_timeout, timeout()}
-                | {allowed_local_paths, [file:name_all()]}
+                | {allowed_local_paths, [binary()]}
                 | {log_warnings, boolean()}
                 | {log_slow_queries, boolean()}
                 | {keep_alive, boolean() | timeout()}
@@ -143,12 +143,13 @@
 %%   </dd>
 %%   <dt>`{connect_timeout, Timeout}'</dt>
 %%   <dd>The maximum time to spend for start_link/1.</dd>
-%%   <dt>`{allowed_local_paths, [file:name_all()]}'</dt>
+%%   <dt>`{allowed_local_paths, [binary()]}'</dt>
 %%   <dd>This option allows you to specify a list of directories or individual
 %%       files on the client machine which the server may request, for example
 %%       when executing a `LOAD DATA LOCAL INFILE' query. Only absolute paths
-%%       are allowed. The default is an empty list, meaning the client will
-%%       not send any local files to the server.</dd>
+%%       without relative components such as `..' and `.' are allowed.
+%%       The default is an empty list, meaning the client will not send any
+%%       local files to the server.</dd>
 %%   <dt>`{log_warnings, boolean()}'</dt>
 %%   <dd>Whether to fetch warnings and log them using error_logger; default
 %%       true.</dd>
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -82,6 +82,7 @@
                 | {database, iodata()}
                 | {connect_mode, synchronous | asynchronous | lazy}
                 | {connect_timeout, timeout()}
+                | {local_files, disallow | allow}
                 | {log_warnings, boolean()}
                 | {log_slow_queries, boolean()}
                 | {keep_alive, boolean() | timeout()}
@@ -142,6 +143,14 @@
 %%   </dd>
 %%   <dt>`{connect_timeout, Timeout}'</dt>
 %%   <dd>The maximum time to spend for start_link/1.</dd>
+%%   <dt>`{local_files, disallow | allow}'</dt>
+%%   <dd>If set to `allow', the client will advertise the `CLIENT_LOCAL_FILES'
+%%       capability in the handshake and send local data on request, for example
+%%       as part of a `LOAD DATA LOCAL INFILE' query. As this is a potential threat
+%%       that would allow hijacked servers to request any file from the client machine
+%%       to which the client has access to, the default is `disallow', so the capability
+%%       is not advertised and any such requests from the server result in an `exit'
+%%       exception with reason `unexpected_local_infile_request'.</dd>
 %%   <dt>`{log_warnings, boolean()}'</dt>
 %%   <dd>Whether to fetch warnings and log them using error_logger; default
 %%       true.</dd>
--- a/src/mysql_protocol.erl
+++ b/src/mysql_protocol.erl
@@ -27,15 +27,13 @@
 %% @private
 -module(mysql_protocol).
 
--export([handshake/9, change_user/8, quit/2, ping/2,
+-export([handshake/8, change_user/8, quit/2, ping/2,
          query/6, fetch_query_response/5, prepare/3, unprepare/3,
          execute/7, fetch_execute_response/5, reset_connnection/2,
          valid_params/1]).
+-export([normalize_split_path/1]).
 
 -type query_filtermap() :: no_filtermap_fun | mysql:query_filtermap_fun().
-
--type local_files_policy() :: allow_local_files
-                            | disallow_local_files.
 
 -type auth_more_data() :: fast_auth_completed
                         | full_auth_requested
@@ -70,21 +68,21 @@
                 Database :: iodata() | undefined,
                 SockModule :: module(), SSLOpts :: list() | undefined,
                 Socket :: term(),
-                LocalFilesPolicy :: local_files_policy(),
                 SetFoundRows :: boolean()) ->
     {ok, #handshake{}, SockModule :: module(), Socket :: term()} |
     #error{}.
+
 handshake(Host, Username, Password, Database, SockModule0, SSLOpts, Socket0,
-          LocalFilesPolicy, SetFoundRows) ->
+          SetFoundRows) ->
     SeqNum0 = 0,
     {ok, HandshakePacket, SeqNum1} = recv_packet(SockModule0, Socket0, SeqNum0),
     case parse_handshake(HandshakePacket) of
         #handshake{} = Handshake ->
             {ok, SockModule, Socket, SeqNum2} =
                 maybe_do_ssl_upgrade(Host, SockModule0, Socket0, SeqNum1, Handshake,
-                                     SSLOpts, Database, LocalFilesPolicy, SetFoundRows),
+                                     SSLOpts, Database, SetFoundRows),
             Response = build_handshake_response(Handshake, Username, Password,
-                                                Database, LocalFilesPolicy, SetFoundRows),
+                                                Database, SetFoundRows),
             {ok, SeqNum3} = send_packet(SockModule, Socket, Response, SeqNum2),
             handshake_finish_or_switch_auth(Handshake, Password, SockModule, Socket,
                                             SeqNum3);
@@ -196,19 +194,19 @@ ping(SockModule, Socket) ->
     {ok, OkPacket, _SeqNum2} = recv_packet(SockModule, Socket, SeqNum1),
     parse_ok_packet(OkPacket).
 
--spec query(Query :: iodata(), module(), term(), local_files_policy(), query_filtermap(),
+-spec query(Query :: iodata(), module(), term(), [binary()], query_filtermap(),
             timeout()) ->
     {ok, [#ok{} | #resultset{} | #error{}]} | {error, timeout}.
-query(Query, SockModule, Socket, LocalFilesPolicy, FilterMap, Timeout) ->
+query(Query, SockModule, Socket, AllowedPaths, FilterMap, Timeout) ->
     Req = <<?COM_QUERY, (iolist_to_binary(Query))/binary>>,
     SeqNum0 = 0,
     {ok, _SeqNum1} = send_packet(SockModule, Socket, Req, SeqNum0),
-    fetch_query_response(SockModule, Socket, LocalFilesPolicy, FilterMap, Timeout).
+    fetch_query_response(SockModule, Socket, AllowedPaths, FilterMap, Timeout).
 
 %% @doc This is used by query/4. If query/4 returns {error, timeout}, this
 %% function can be called to retry to fetch the results of the query.
-fetch_query_response(SockModule, Socket, LocalFilesPolicy, FilterMap, Timeout) ->
-    fetch_response(SockModule, Socket, Timeout, text, LocalFilesPolicy, FilterMap, []).
+fetch_query_response(SockModule, Socket, AllowedPaths, FilterMap, Timeout) ->
+    fetch_response(SockModule, Socket, Timeout, text, AllowedPaths, FilterMap, []).
 
 %% @doc Prepares a statement.
 -spec prepare(iodata(), module(), term()) -> #error{} | #prepared{}.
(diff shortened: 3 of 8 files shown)


juhlig:
I revised my initial work a bit, as I realized that unconditionally enabling this capability would enable a MySQL server (or something _impersonating_ a MySQL server) to trick the client into sending the contents of arbitrary files that it has access to.

For this purpose, the `local_files` option is introduced here, with the possible values being `disallow` (the default) or `allow`. First, it controls if the `CLIENT_LOCAL_FILES` capability flag is sent to the server in the handshake. As not setting the flag does not prevent a rogue server from still sending requests for local files, `disallow` also prevents the client from honouring such requests, and fails with an `exit` exception with reason `unexpected_local_infile_request` if such a packet is received.

zuiderkwast:
Nice to see this old feature request revived. I hope someone actually needs this feature. :-)

It's interesting that even if the client specifies the filename, the server can request any filename it wants, a security issue as you pointed out. I'm thinking that the client should only send the file that was has specified in the `LOAD LOCAL DATA INFILE` query, but I realize that is not possible since we would need to parse SQL to do that.

Using an option is a good idea, but it is unfortunate that the client has to `allow` access to the entire filesystem to use this feature.

When I first saw the option name `local_files`, I assumed that the value is a list of filenames (because the name is in plural) which is isn't, but maybe it can be? It's an idea anyway. Examples:

```Erlang
    {local_files, ["/"]}                        % Allow access to all files in the
                                                % filesystem. The database is trusted.
```
```Erlang
    {local_files, ["/path/to/stuff",            % Allow sending data files under
                   "/path/to/other/stuff"]}     % specific directories.
```

juhlig:
> Nice to see this old feature request revived. I hope someone actually needs this feature. :-)

even if not, it makes the client more complete :)

> It's interesting that even if the client specifies the filename, the server can request any filename it wants, a security issue as you pointed out.

Yeah... the server may always send something like `<<16#fb, "/path/to/file">>`... I wonder if other clients (Erlang or any other) are prepared for this.

> I'm thinking that the client should only send the file that was has specified in the `LOAD LOCAL DATA INFILE` query, but I realize that is not possible since we would need to parse SQL to do that.

Yeah, and I sure don't want to go down that path ;)

> Using an option is a good idea, but it is unfortunate that the client has to `allow` access to the entire filesystem to use this feature.

Hm, point taken :)

> When I first saw the option name `local_files`,

I just picked that name because the capabilities flag is named that way.

> I assumed that the value is a list of filenames (because the name is in plural) which is isn't, but maybe it can be? It's an idea anyway. Examples:
> 
> ```erlang
>     {local_files, ["/"]}                        % Allow access to all files in the
>                                                 % filesystem. The database is trusted.
> ```
> 
> ```erlang
>     {local_files, ["/path/to/stuff",            % Allow sending data files under
>                    "/path/to/other/stuff"]}     % specific directories.
> ```

Hm, good idea. Not sure a list of paths (vs a single path) is needed, but it is not that big a hassle to implement it that way, I guess :) I will have to think up some path validation function for this, though, especially if we consider relative paths. While there is a `safe_relative_path` function that may help here, it only appears in OTP/19.3, and got deprecated again and moved to another module in OTP/23... Well, I'll think of something :)

zuiderkwast:
> Not sure a list of paths (vs a single path) is needed

Right, a single path is enough, e.g. `{local_files_dir, "/path/to/dir"}`. For relative paths, it's better if it's a single path.

Maybe simplistic, but can't we use `filename:absname_join(LocalFilesDir, RelPath)` to resolve relative paths and then check if the result is inside the LocalFilesDir by simply checking if it's a prefix?

juhlig:
> > Not sure a list of paths (vs a single path) is needed
> 
> Right, a single path is enough, e.g. `{local_files_dir, "/path/to/dir"}`. For relative paths, it's better if it's a single path.

I was thinking more along the lines of it meaning "this path is safe to send files from, don't send files from other paths". And in reconsideration, this _would_ justify a list of paths instead of a single one...

In contrast, I think what you imagine is a "root path" for local files? I have given this some consideration, but I think I came to the conclusion that this is quite complex. To make it a true root path, we would have to make it that for absolute paths as well, not only relative (meaning `"/foo/bar"` would really mean `LocalFilesDir ++ "foo/bar"`). In turn, this would make relative paths pointless (both `"/foo/bar"` and `"foo/bar"` would be the same thing, `LocalFilesDir ++ "`foo/bar"`), unless we also take the current working directory into consideration, making it even more complex. It might be confusing to users, too, even given proper documentation.

> Maybe simplistic, but can't we use `filename:absname_join(LocalFilesDir, RelPath)` to resolve relative paths and then check if the result is inside the LocalFilesDir by simply checking if it's a prefix?

Relative paths are not resolved by this function (or any other I know of, not even `absname/1,2`). Meaning, `filename:absname_join("/foo/", "../bar")` would result in `"/foo/../bar"`, prefix-matching `"/foo/"` while it really points at `/bar` :\

zuiderkwast:
Seems there are multiple options for this little thing:

1. Use a list of safe paths. (Unclear what's the base for relative paths. Current directory of the process?)
2. Use a single safe path, which also can serve as the base directory if relative paths are provided. Absolute paths must have this safe path as a prefix.
3. Use a root directory, which also servers as the base for relative paths, effectively making "/foo" and "foo" equivalent.

To circumvent the problem with unsafe paths, can we just forbid ".." to occur in the paths?

Number 2 might be less confusing than 3 for users, but who knows, maybe it's the other way around... Choosing a good name of the option is probably important.

juhlig:
> Number 2 might be less confusing than 3 for users, but who knows, maybe it's the other way around... Choosing a good name of the option is probably important.

I'm actually in favor of option 1, maximum flexibility, little confusion (IMO), and the choice of the current working directory for relative paths seems the natural thing to do.

In fact, this is what I did with the last commit, which I was already experimenting with when you commented ;) Tests for non-allowed paths are still missing, but you may check it out already.

To address the issue with requests for forbidden or non-existing files, I introduced an extra field in the `#error{}` record, to indicate that it is a client-side error. That is just a hack for now, but I haven't come up with a better idea ;)

juhlig:
Still fiddling with the path resolution etc.

I also stumbled upon the reason why none of the `filename` functions does any path resolving/normalization. It has to do with symlinks: Assume a directory structure like this on a filesystem that supports symlinks:
```
/
|
+-- dir1/
|   |
|   +-- dir1.1/
|
+-- dir2/
    |
    +-- dir2.1/ (symlink to dir1.1)
```
A path like `/dir2/dir2.1/..` will then land you in `/dir1` on a real file system. The `filename` module does not work with real filesystems, however, only with hypothetical filenames, so it doesn't know symlinks etc, and thus a resolve/normalize function in that module would resolve the same path to `/dir2`. And this is probably the reason why they moved the `safe_relative_path` function from that `filename` over to `filelib` (which _does_ work with real filesystems) when they included a check for symlinks...

zuiderkwast:
Interesting. But can't we simply forbid ".." in the paths? It is really simple and no real drawback.

Jan Uhlig <notifications@github.com> skrev: (8 juli 2020 16:46:39 CEST)

juhlig:
I guess i'm some kind of perfectionist sometimes ^^;;;

src/mysql_protocol.erl
```diff
@@ -196,26 +193,19 @@ ping(SockModule, Socket) ->
...
-    {ok, [#ok{} | #resultset{} | #error{}]} | {error, timeout}.
-query(Query, SockModule, Socket, Timeout) ->
-    query(Query, SockModule, Socket, no_filtermap_fun, Timeout).
-
--spec query(Query :: iodata(), module(), term(), query_filtermap(), timeout()) ->
+-spec query(Query :: iodata(), module(), term(), [binary()], query_filtermap(),
+            timeout()) ->
     {ok, [#ok{} | #resultset{} | #error{}]} | {error, timeout}.
-query(Query, SockModule, Socket, FilterMap, Timeout) ->
+query(Query, SockModule, Socket, AllowedPaths, FilterMap, Timeout) ->
```
zuiderkwast, line 199: These parameters are close to many. (If we add more, we could perhaps wrap them in a record...)

src/mysql_protocol.erl
```diff
@@ -1220,6 +1214,96 @@ recv_packet(SockModule, Socket, Timeout, ExpectSeqNum, Acc) ->
...
+normalize_split_path(Root, [Segment|More], Acc) ->
+    normalize_split_path(Root, More, [Segment|Acc]).
+
+-spec send_file_chunk(module(), term(), Handle :: file:io_device(), SeqNum :: integer()) ->
+    {ok, NextSeqNum :: integer()}.
+send_file_chunk(SockModule, Socket, Handle, SeqNum0) ->
+    case file:read(Handle, 16#ffffff) of
+        eof ->
+            {ok, SeqNum0};
+        {ok, <<>>} ->
```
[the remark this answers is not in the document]
juhlig: That's a precautionary thing. While I have never seen it happen, the documentation does not say that it _can't _ return empty. And if it does return an empty binary but it is not on eof, it will send an empty packet to the server, indicating eof to the server while the file has actually not been fully sent.

juhlig:
>     1. I don't think the benefit of the fancy path resolving logic is worth the added complexity. Besides, it might not be correct, given symlinks. What's the benefit of supporting ".." in paths anyway? I can't see any really.

For one, I'm still in an exploratory phase here. If we can make it work with "anything", I think we should. OTOH, it _is_ getting kinda complex, and still not working with symlinks. For that to work, we would have to work on the actual filesystem and resolve symlinks, much like `filelib:safe_relative_path/2` does. As this does add to the complexity, I'm (almost ;)) with you that we should just forbid `..`.

> If you come up with a use case, I may change my mind.

Not sure if it is valid enough (so I'm just pointing it out), but for one it may be a thing with programmatically constructed paths, ie not user created ones (dunno if I made myself clear, my english is not yet up to speed at this early hour ^^;). Another thing is when you use a relative path, want to stay in the CWD but point to a location above/outside it.

>        We can simply document that paths in SELECT LOCAL INFILE can't contain ".." and then we have a very simple validation logic.
> 
>     2. The validation error. The external API is what's important here. The internals can always be changed later. We could perhaps use an unused error code (negative number or zero?) and document that

That seemed somewhat a lazy/half-assed way to do it ^^; But sure, if it's ok with you, I can do that.

> so it fits the pattern of the other errors. Or we can extend the `query_result()` to include `{error, client_reason()}` also.

Problem I had there was that it would be indistinguishable from server errors then.

> Another option would be to raise a `badarg` error rather than returning an error if path validation fails, because we can assume that the server only requests paths which are given in the query or parameters.

If we do that at this point, the connection process will crash, which is probably not really the best thing. Only the query is wrong, as you gave a path in `LOAD LOCAL INFILE`) that is outside your allowed paths (or a file that doesn't exist or sth). The server reacted in a proper way, the connection is really still usable (given that we do send _something_ like an empty packet).

>        Btw, I see that the error tuple is not documented anywhere. It could be documented at `query/5` for example or in the general section for the module.

It is really still just a hack, I only there so I can progress ;)

src/mysql_protocol.erl
```diff
@@ -1220,6 +1214,96 @@ recv_packet(SockModule, Socket, Timeout, ExpectSeqNum, Acc) ->
...
+
+normalize_split_path(Root, [], Acc) ->
+    [Root|lists:reverse(Acc)];
+normalize_split_path(Root, ["."|More], Acc) ->
+    normalize_split_path(Root, More, Acc);
+normalize_split_path(Root, [<<".">>|More], Acc) ->
+    normalize_split_path(Root, More, Acc);
+normalize_split_path(Root, [".."|More], []) ->
+    normalize_split_path(Root, More, []);
+normalize_split_path(Root, [<<"..">>|More], []) ->
```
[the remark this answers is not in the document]
juhlig: We are already at the top of the path (`[]` in the `Acc` param), can't go any further up. The functions in `file` etc just ignore that, also, so I went with that instead of erroring.

src/mysql_protocol.erl
```diff
@@ -1220,6 +1214,96 @@ recv_packet(SockModule, Socket, Timeout, ExpectSeqNum, Acc) ->
...
+    [Root|RelPath] = filename:split(Path2),
+    normalize_split_path(Root, RelPath, []).
+
+normalize_split_path(Root, [], Acc) ->
+    [Root|lists:reverse(Acc)];
+normalize_split_path(Root, ["."|More], Acc) ->
+    normalize_split_path(Root, More, Acc);
+normalize_split_path(Root, [<<".">>|More], Acc) ->
+    normalize_split_path(Root, More, Acc);
+normalize_split_path(Root, [".."|More], []) ->
```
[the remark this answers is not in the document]
juhlig: Yes, but it is used on the allowed paths also, and that can be lists, too.

src/mysql_protocol.erl
```diff
@@ -1220,6 +1214,96 @@ recv_packet(SockModule, Socket, Timeout, ExpectSeqNum, Acc) ->
...
+    normalize_split_path(Root, More, []);
+normalize_split_path(Root, [<<"..">>|More], []) ->
+    normalize_split_path(Root, More, []);
+normalize_split_path(Root, [".."|More], [_|Acc]) ->
+    normalize_split_path(Root, More, Acc);
+normalize_split_path(Root, [<<"..">>|More], [_|Acc]) ->
+    normalize_split_path(Root, More, Acc);
```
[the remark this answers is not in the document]
juhlig: Yes. As pointed out elsewhere, it may be possible to work on the real file system (like `filelib:safe_relative_path/2`, we need to get at the real file in the next step, anyway), but ignoring `..` is really just simpler, so we may end up with that, anyway.

src/mysql_protocol.erl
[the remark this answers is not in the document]
zuiderkwast: Because the root is the parent of itself, for historical reasons. I found this on the Wikipedia article on inode:

*"Historically, it was possible to hard link directories. This made the directory structure into an arbitrary directed graph contrary to a directed acyclic graph. It was even possible for a directory to be its own parent. Modern systems generally prohibit this confusing state, except that the parent of root is still defined as root. The most notable exception to this prohibition is found in Mac OS X (versions 10.5 and higher) which allows hard links of directories to be created by the superuser."*

I wonder what trouble hard links can cause if used as in your example with `/dir2/dir2.1/..` on Mac OS X...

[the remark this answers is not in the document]
juhlig: No idea... but we had our share with MacOS just recently in `ranch` ^^; Trouble is, I don't have any MacOS at hand, so I only see that it doesn't pass when it gets into the CI... ^^;;;

I'll get back to this PR next week I guess, and I'm now of a mind to just disallow `..` in paths, as you suggest.

zuiderkwast:
OK, so will go with forbidding ".." then, good! For programmatically constructed paths, the ".." can be substituted programmatically too, pushing the problem to the application.

Regarding the error, if we raise error(badarg) from mysql:query/N, it doesn't mean we have to crash the connection process. Just want to point that out.

If we introduce client_error(), it is distinguishable from a server_error() if it's not a 3-tuple. It does clutter the interface though, slightly. If this is our only client error, it's not justified IMO. We do have {error, timeout} in some cases though, don't we?

The 3rd option is encoding it as a server error tuple with some unused error code. I think it's slightly simpler (better) than the 2nd option, perhaps, since it doesn't add new structure.

juhlig:
> OK, so will go with forbidding ".." then, good! For programmatically constructed paths, the ".." can be substituted programmatically too, pushing the problem to the application.

I tried, but it still turns out complex, which means there are probably edge cases where it could be tricked... So I went back to the beginning and thought this through again. This is what I came up with, lets see what you think.

* we only allow binaries as paths. what we get from the server is a binary, matching that against maybe-utf8 string (or iodata) allowed paths is hard, and probably not worth it
* we only allow _absolute_ paths (or volumerelative ones that are not relative to the current volume, ie only ones that have a drive letter in them). this makes validation easier, and is maybe even less confusing for users than if we were considering CWD and relative paths.
* we do not allow and `..` or `.` as path segments
* allowed paths can either be specific files, or directories. in the latter case, they must end with a `/`
* with these restrictions, we can do a match with the allowed and requested paths like this:
  * if the requested path exactly matches a specific allowed file (an allowed path not ending with a `/`), it is allowed and sent
  * if the requested path prefix-matches an allowed directory (an allowed path ending with a `/`), ie the allowed path is a prefix of the requested path, it is allowed and sent
  * everything else is forbidden and not sent

This ruleset may seem quite confusing, but actually it is really simple and straight-forward. The rules are restrictive, but simple and clear. What do you think?

Regarding the errors, (unused error code, like negative or sth) I would go with the third option then.

juhlig:
Ok, with the last commit, the path checking works as described above.

Regarding the error, it now returns `-1` as error code, and the error tuple in the `msg` field of the record (up for discussion ;)).

include/records.hrl
```diff
@@ -40,7 +40,7 @@
              msg :: binary()}).
 %% Error packet, commonly used in the protocol.
 -record(error, {code :: integer(), state :: binary() | undefined,
-                msg :: binary()}).
+                msg :: binary() | {error, term()}}).
```
zuiderkwast, line 43: So the new error comes out as code =  -1, state = undefined, msg = `{error, not allowed}`. Not very clear. (What's not allowed?) Also the tuple might break existing user code doing things like logging the errors and assuming msg is binary. I think msg should be a descriptive sentence as for the server errors. Idea: "Server requested a file not permitted by the client: /etc/passwd". Possibly truncating the file name if > 100 chars or so. The code -1 is fine and undefined state s not new so also fine.

src/mysql_protocol.erl
```diff
@@ -1220,6 +1214,86 @@ recv_packet(SockModule, Socket, Timeout, ExpectSeqNum, Acc) ->
             {error, Reason}
     end.
 
+-spec send_file(module(), term(), Filename :: binary(), AllowedPaths :: [binary()],
+                SeqNum :: integer()) ->
+    {ok | {error, term()}, NextSeqNum :: integer()}.
```
zuiderkwast, line 1219: Better be explicit and say `{error, not_allowed}` in the spec here.

src/mysql_protocol.erl
```diff
@@ -1220,6 +1214,86 @@ recv_packet(SockModule, Socket, Timeout, ExpectSeqNum, Acc) ->
...
+-spec allowed_path(binary(), [binary()]) -> boolean().
+allowed_path(Path, AllowedPaths) ->
+    valid_path(Path) andalso
+    binary:last(Path) =/= $/ andalso
+    lists:any(
+        fun
+            (AllowedPath) when Path =:= AllowedPath ->
+                true;
+            (AllowedPath) when binary_part(Path, 0, byte_size(AllowedPath)) =:= AllowedPath ->
+                binary:last(AllowedPath) =:= $/;
```
zuiderkwast, line 1244: I think it's a bit harsh to require directories to he specified with a trailing slash. It's not required by most unix commands and applications in general. It wouldn't be too hard to add something like a fun clause for allowed directory without trailing slash: `(<<AllowedPath:Size/binary, "/", _/binary>>) when AllowedPath doesn't end with a slash`.

src/mysql.erl
```diff
@@ -142,6 +143,16 @@
 %%   </dd>
 %%   <dt>`{connect_timeout, Timeout}'</dt>
 %%   <dd>The maximum time to spend for start_link/1.</dd>
+%%   <dt>`{allowed_local_paths, [binary()]}'</dt>
+%%   <dd>This option allows you to specify a list of directories or individual
+%%       files on the client machine which the server may request, for example
+%%       when executing a `LOAD DATA LOCAL INFILE' query. Only absolute paths
+%%       without relative components such as `..' and `.' are allowed.
+%%       A path ending with a `/' denotes a directory from which files
```
zuiderkwast, line 151: If we handle dirs with or without slash, we don't need this sentence. No need to distinguish between dirs and regular files.

zuiderkwast (changes requested):
Sorry for the delay! I have a few comments on small things. Then I'd be happy to have this merged.

include/records.hrl
juhlig: Ok, changed as requested. There are now 2 error codes, though, `-1` for not allowed, `-2` for error opening file (more on that below).

src/mysql_protocol.erl
juhlig: `not_allowed` is not the only error that could be returned from this function, but also any error that is returned from `file:open/2`.

src/mysql_protocol.erl
juhlig: Ok, I changed as suggested, to work with allowed paths with and without slashes.

src/mysql.erl
juhlig: ... and removed the sentence.

juhlig:
The tests fail, I'll have a look at that later. They pass on my local machine, but there I was using OTP/23, so probably something that works only there.

juhlig:
Tests pass now :) I'll remove the WIP from the title then.

zuiderkwast:
Ah, right. But I thought the type of these were defined as
`file:posix()` or something like that.

zuiderkwast (approved):

juhlig:
> Ah, right. But I thought the type of these were defined as `file:posix()` or something like that.

Ok, I changed the spec to be more elborate concerning the error reason to `not_allowed` and everything `file:open` might return as error reason.

zuiderkwast:
Awesome! It's a bit hard to see what's changed in the last change when force-pushing. It's easier to keep track if we use multiple commits during development and squash them when (or just before) submitting. Not a big deal though. I see the path check code looks really nice and simple now.

juhlig:
Ok, so as far as I'm concerned, I think this is ready :)

PR #170, opened 2020-12-03
README: markdown syntax highlighting

szTheory:


zuiderkwast (commented):
Good, but you should also remove 4 spaces of indentation.

szTheory:
Thanks, updated! I originally didn't indent it because some people like a smaller diff

zuiderkwast (approved):
Awesome!

There is a problem (not your fault) in the build which seems to be related to MariaDB handling of a SLEEP which is interrupted. Something seems to have changed in the build environment. I created #171 for this. Once it's fixed and the builds are green again, we can merge this PR.

PR #172, opened 2021-01-18
allow prepare query to have a timeout

neeraj9:
It's been a while but this patch was sitting in
one of my private gist. I believe the issue was
during one of the usage of mysql wherein the underlying
assumption was that timeout was applied but eventaully
there were a lot of prepare queries which ended up
in locking the table. Hence the application started
experiencing table lock errors when there were too many
prepare queries and MySQL was unable to cope up with it.
This patch attempts to fix such scenarios and fail with
timeout being applied.

### DIFF
--- a/src/mysql_protocol.erl
+++ b/src/mysql_protocol.erl
@@ -28,7 +28,7 @@
 -module(mysql_protocol).
 
 -export([handshake/8, change_user/8, quit/2, ping/2,
-         query/6, fetch_query_response/5, prepare/3, unprepare/3,
+         query/6, fetch_query_response/5, prepare/3, unprepare/3, prepare/4,
          execute/7, fetch_execute_response/5, reset_connnection/2,
          valid_params/1, valid_path/1]).
 
@@ -210,9 +210,19 @@ fetch_query_response(SockModule, Socket, AllowedPaths, FilterMap, Timeout) ->
 %% @doc Prepares a statement.
 -spec prepare(iodata(), module(), term()) -> #error{} | #prepared{}.
 prepare(Query, SockModule, Socket) ->
+    prepare(Query, SockModule, Socket, infinity).
+
+%% @doc Prepares a statement.
+-spec prepare(iodata(), module(), term(), infinity | timeout()) -> #error{} | #prepared{}.
+prepare(Query, SockModule, Socket, Timeout) ->
     Req = <<?COM_STMT_PREPARE, (iolist_to_binary(Query))/binary>>,
     {ok, SeqNum1} = send_packet(SockModule, Socket, Req, 0),
-    {ok, Resp, SeqNum2} = recv_packet(SockModule, Socket, SeqNum1),
+    {ok, Resp, SeqNum2} = case recv_packet(SockModule, Socket, Timeout, SeqNum1) of
+                              {error, timeout} ->
+                                  %% drastic measure, so kill the connection
+                                  _ = query(<<"kill connection connection_id()">>, SockModule, Socket, no_filtermap_fun, Timeout),
+                               Result -> Result
+                          end,
     case Resp of
         ?error_pattern ->
             parse_error_packet(Resp);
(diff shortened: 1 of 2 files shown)


src/mysql_protocol.erl
```diff
@@ -210,9 +210,19 @@ fetch_query_response(SockModule, Socket, AllowedPaths, FilterMap, Timeout) ->
 %% @doc Prepares a statement.
 -spec prepare(iodata(), module(), term()) -> #error{} | #prepared{}.
 prepare(Query, SockModule, Socket) ->
+    prepare(Query, SockModule, Socket, infinity).
```
zuiderkwast, line 213: Have you considered changing this to a hard-coded value such as `?cmd_timeout`, which is used for BEGIN, COMMIT and ROLLBACK? Do you think it would break backward compatibility to do that?

src/mysql_protocol.erl
```diff
@@ -28,7 +28,7 @@
 -module(mysql_protocol).
 
 -export([handshake/8, change_user/8, quit/2, ping/2,
-         query/6, fetch_query_response/5, prepare/3, unprepare/3,
+         query/6, fetch_query_response/5, prepare/3, unprepare/3, prepare/4,
```
zuiderkwast, line 31: Please delete prepare/3 and change all the calls to use prepare/4 instead.

src/mysql_protocol.erl
```diff
@@ -210,9 +210,19 @@ fetch_query_response(SockModule, Socket, AllowedPaths, FilterMap, Timeout) ->
...
+%% @doc Prepares a statement.
+-spec prepare(iodata(), module(), term(), infinity | timeout()) -> #error{} | #prepared{}.
+prepare(Query, SockModule, Socket, Timeout) ->
     Req = <<?COM_STMT_PREPARE, (iolist_to_binary(Query))/binary>>,
     {ok, SeqNum1} = send_packet(SockModule, Socket, Req, 0),
-    {ok, Resp, SeqNum2} = recv_packet(SockModule, Socket, SeqNum1),
+    {ok, Resp, SeqNum2} = case recv_packet(SockModule, Socket, Timeout, SeqNum1) of
+                              {error, timeout} ->
+                                  %% drastic measure, so kill the connection
+                                  _ = query(<<"kill connection connection_id()">>, SockModule, Socket, no_filtermap_fun, Timeout),
```
zuiderkwast, line 223: Keep line length 80 please.

zuiderkwast (commented):
Thank you for reporting this and for the fix!

I'm reluctant to adding one more variant if we can solve it by setting a fixed timeout on prepare instead.

Do you know why prepare takes a long time? Does it need to look up indexes and things like that, like a regular query?

neeraj9:
> Thank you for reporting this and for the fix!
> 
> I'm reluctant to adding one more variant if we can solve it by setting a fixed timeout on prepare instead.
> 
> Do you know why prepare takes a long time? Does it need to look up indexes and things like that, like a regular query?

One of the reason for longer Prepare is due to running a lot of them on the same database. I believe there is an inherent optimization to remove excessive use of Prepare statements, but if that happens (even momentarily) then we should be covered there.

src/mysql_protocol.erl
neeraj9: ah! This would break existing applications which depend on prepare/3. I am not sure if that would be a great idea.

src/mysql_protocol.erl
neeraj9: Sure. Sounds good.

src/mysql_protocol.erl
neeraj9: I am not sure if applications have hard dependencies there. Thinking through this again, maybe we can get away with breaking things by announcing it under some page easily visible (for changelog) and then leaving it for others to find during their testing if you did blind upgrades.

src/mysql_protocol.erl
zuiderkwast: In `mysql` we must keep prepare/3 in of course, but `mysql_protocol` is an internal module. We have removed arity variants here before, not long ago. That's why there is only query/6 for query for example.

src/mysql_protocol.erl
zuiderkwast: Yes, it's an interesting question. If we set a quite high number such as 10 seconds, I don't think it will break any applications. It will timeout only if the database is overloaded, like in your case, and then it's good that it times out.

How many seconds are you planning to use for your application? Not yet decided?

Gemrav (commented):
G

PR #176, opened 2021-06-29
Add optional float-as-decimal encoding

### LINKED ISSUE #52: Problems with decimal columns
Hello,

I have the following test schema:

``` sql
CREATE TABLE `test_decimals` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `balance` decimal(13,4) NOT NULL DEFAULT '0.0000',
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=15 DEFAULT CHARSET=utf8;

INSERT INTO test_decimals(id, balance) VALUES(1, 5000);
```

And the following code snippet:

``` erlang
-module(test_decimal).

-export([start/0, exec/1]).

start() ->
    {ok, Pid} = mysql:start_link([
        {host, "127.0.0.1"},
        {port, 3306},
        {user, "root"},
        {password, "root"},
        {database, "test_db"},
        {keep_alive, 300000},
        {queries, [<<"set names 'utf8';">>]},
        {prepare, [{debit_account, <<"UPDATE test_decimals SET balance = balance - ? WHERE id = ?">>}]}
    ]),

    Pid.

exec(Pid)->
    mysql:execute(Pid, debit_account, [10.2, 1]).
```

If you run `Pid = test_decimal:start().` to create a connection and then run several times `test_decimal:exec(Pid).` you will notice from time to time the following warning:

``` erlang
[warning] Note 1265: Data truncated for column 'balance' at row 1
 in UPDATE test_decimals SET balance = balance - ? WHERE id = ?
```

I tested the same query with `mysql workbench`, `emysql` client, and also the mysql console and I cannot see this warning.

I'm sure is related on how floats are encoded in `mysql-otp` in prepared statements. 

Any Idea how I can get rid of this problem ?

Silviu

zuiderkwast:
* Possibility to send numbers as decimals using `{decimal, Value}`.
* New option `{float_as_decimal, boolean() | non_neg_integer()}` to
  do this automatically for floats, optionally rounded to a given
  number of decimals.

Fixes #52


zuiderkwast:
@silviucpp I included test cases to reproduce the warnings you mentioned in #52. Please have a look!

silviucpp:
Looks great !

src/mysql_encode.erl
```diff
@@ -23,6 +23,12 @@ encode(String) when is_list(String) ->
     encode(unicode:characters_to_binary(String));
 encode(Bitstring) when is_bitstring(Bitstring) ->
     ["b'", [ case B of 0 -> $0; 1 -> $1 end || <<B:1>> <= Bitstring ], $'];
+encode({decimal, Num}) when is_float(Num); is_integer(Num) ->
+    encode(Num);
+encode({decimal, Str}) when is_binary(Str); is_list(Str) ->
+    %% Simple injection block
+    nomatch = re:run(Str, <<"[^0-9\\.eE-]">>),
```
zuiderkwast, line 30: `+` needs to be accepted too, as in `1.0+e10`

src/mysql_encode.erl
```diff
@@ -23,6 +23,12 @@ encode(String) when is_list(String) ->
     encode(unicode:characters_to_binary(String));
 encode(Bitstring) when is_bitstring(Bitstring) ->
     ["b'", [ case B of 0 -> $0; 1 -> $1 end || <<B:1>> <= Bitstring ], $'];
+encode({decimal, Num}) when is_float(Num); is_integer(Num) ->
+    encode(Num);
+encode({decimal, Str}) when is_binary(Str); is_list(Str) ->
+    %% Simple injection block
+    nomatch = re:run(Str, <<"[^0-9\\.eE+-]">>),
```
Maria-12648430, line 30: The `.` loses its special meaning when used in a character class, no need to escape it.

Also, while it is unlikely that this regexp ever changes, it might be good to add a comment to remember that the `-` must be the last character in a character class, otherwise it becomes a range operator. And/or escape the `-` just in case.

src/mysql_protocol.erl
```diff
@@ -1078,6 +1078,18 @@ encode_param(Value) when is_integer(Value), Value < 0 ->
...
+                  Value;
+              is_list(Value) ->
+                  list_to_binary(Value);
+              is_integer(Value) ->
+                  integer_to_binary(Value);
+              is_float(Value) ->
+                  list_to_binary(io_lib:format("~w", [Value]))
+          end,
+    EncLength = lenenc_int_encode(byte_size(Bin)),
+    {<<?TYPE_DECIMAL, 0>>, <<EncLength/binary, Bin/binary>>};
```
Maria-12648430, line 1092: Alternatively, maybe do it with multiple function clauses like this?

```erlang
...
encode_param({decimal, Bin}) when is_binary(Bin) ->
    EncLength = lenenc_int_encode(byte_size(Bin)),
    {<<?TYPE_DECIMAL, 0>>, <<EncLength/binary, Bin/binary>>};
encode_param({decimal, List}) when is_list(List) ->
    encode_param({decimal, list_to_binary(List)});
...
```

In case of an invalid type for `Value`, this would result in a `function_clause` instead of an `if_clause` error.

Maria-12648430 (approved):
Approved, just minor comments in the code ;)

src/mysql_encode.erl
zuiderkwast: Good ideas! I will change.

src/mysql_protocol.erl
zuiderkwast: It looks clean but re-wrapping it in a tuple again is unnecessary work. It doesn't matter which error it crashes with. (Perhaps I can just skip the likebreaks after `->` in the if clauses so it looks more compact?)

Maria-12648430: In that case, I'd leave it as it is. Removing the linebreaks would make it hard to read IMO.

juhlig (approved):
Looks good to me, nothing to add or change :)

zuiderkwast:
Thanks for looking!

Maria-12648430:
You're welcome :)

PR #178, opened 2021-08-02
Add auth_plugin_name to the end of handshake response 

suexcxine:
according to mysql specs
https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::HandshakeResponse

The motivation is that, .. some mysql proxy reports error(Bad handshake) when I try to connect using mysql-otp
And I tested with myxql and it's ok
So I compared every bits of handshake response sent by mysql-otp and myxql
And discovered the difference


zuiderkwast:
Good catch!

Issue #180, opened 2021-08-31
CI: Switch to GitHub Actions (or something else)


zuiderkwast:
travic-ci.org has stopped working. We need to switch to travis-ci.com, or perhaps to GitHub Actions or something else.

zuiderkwast:
This is kind of urgent. I don't want to merge new stuff without a working CI.

I've tried moving the jobs from travis-ci.org to travis-ci.com, which seems easy for personal repos but not organisations. @juhlig?

zuiderkwast:
This one looks promising: https://github.com/marketplace/actions/actions-setup-mysql

> Available Versions are:
>
> * MySQL
>     *   8.0
>     *   5.7
>     *   5.6
> *   MariaDB
>     *   10.6
>     *   10.5
>     *   10.4
>     *   10.3

juhlig:
> This one looks promising: https://github.com/marketplace/actions/actions-setup-mysql

Yes :) If we can't get it working, we could also ask the maintainer for help I guess.

zuiderkwast:
@Gemrav that's great! After two years, there are probably more and easier ways to install different versions of these databases in github actions.

juhlig:
Fixed by #207.

PR #181, opened 2021-09-01
with unix domain socket {local, /tmp/mysql.sock} used, disable 'nodelay' tcp option

dogtapdsd:
disable {nodelay, bool()} when use domain = local access mysql.

### DIFF
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -158,8 +158,14 @@ connect(#state{connect_timeout = ConnectTimeout} = State) ->
     end.
 
 connect_socket(#state{tcp_opts = TcpOpts, host = Host, port = Port} = State) ->
+    %% 'nodelay' option is not supported for domain = local, but if inet_backend =/= socket this error will be ignored.
+    CanSetNodelay = case Host of
+        {local, _LocalAddr} -> false;
+        _NonLocalAddr -> true
+    end,
+
     %% Connect socket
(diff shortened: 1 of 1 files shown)


src/mysql_conn.erl
```diff
@@ -158,8 +158,14 @@ connect(#state{connect_timeout = ConnectTimeout} = State) ->
     end.
 
 connect_socket(#state{tcp_opts = TcpOpts, host = Host, port = Port} = State) ->
+    %% 'nodelay' option is not supported for domain = local, but if inet_backend =/= socket this error will be ignored.
+    CanSetNodelay = case Host of
+        {local, _LocalAddr} -> false;
+        _NonLocalAddr -> true
+    end,
```
zuiderkwast, line 165: Instead of a boolean, we can use meaningful values, for example:

```suggestion
    Transport = case Host of
        {local, _LocalAddr} -> local;
        _NonLocalAddr -> tcp
    end,
```

zuiderkwast (commented):
Looks good in general, but it's a problem that the CI jobs are not running, because travis-ci.org has stopped working. See #180. I don't really like to merge stuff without a working CI.

src/mysql_conn.erl
dogtapdsd: Indeed，this modifications have more readability

juhlig:
Replaced by #184

PR #184, opened 2021-12-06
Remove and do not auto-add TCP option nodelay if not supported

### LINKED ISSUE #179: using otp 24 with inet_backend = socket , [{host, {local, "/tmp/mysql.sock"}},{tcp_options,  [{nodelay,Bool}] Option lead to {error, enotsup} Error
Using unix domain socket '{local, "/tmp/mysql.sock"}' to connect mysql will case this issuse.
the otp gen_tcp document explain this compatibility problem
[https://erlang.org/doc/man/gen_tcp.html](url)

> 
The option nodelay is a TCP specific option that is not compatible with domain = local.
When using inet_backend = socket, trying to create a socket (via listen or connect) with domain = local (for example with option {ifaddr, {local,"/tmp/test"}}) will fail with {error, enotsup}.
This does not actually work for inet_backend = inet either, but in that case the error is simply ignored, which is a bad idea. We have choosen to not ignore this error for inet_backend = socket.

juhlig:
Closes #179. Replaces #181.

### DIFF
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -159,7 +159,7 @@ connect(#state{connect_timeout = ConnectTimeout} = State) ->
 
 connect_socket(#state{tcp_opts = TcpOpts, host = Host, port = Port} = State) ->
     %% Connect socket
-    SockOpts = sanitize_tcp_opts(TcpOpts),
+    SockOpts = sanitize_tcp_opts(Host, TcpOpts),
     {ok, Socket} = gen_tcp:connect(Host, Port, SockOpts),
 
     %% If buffer wasn't specifically defined make it at least as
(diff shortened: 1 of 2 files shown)


zuiderkwast (approved):

zuiderkwast:
Unfortunately we don't have a CI at the moment.

zuiderkwast:
@juhlig if you run the tests locally (preferably with different versions of MariaDB and MySQL and OTP) feel free to merge...

juhlig:
Oh, I didn't notice there already was a PR... =^^=

One thing came to mind, though: The way both this and the other PR work, they will remove `nodelay` even if it was explicitly given in the TCP options by the user, instead of just not auto-adding it. Should we do this, or let the error occur if the user explicitly "asked for it"?

As for CI... I can test with different OTP versions, but not with different MySQL/MariaDB versions. I'll see if @Maria-12648430 can help ;)

zuiderkwast:
> One thing came to mind, though: The way both this and the other PR work, they will remove `nodelay` even if it was explicitly given in the TCP options by the user, instead of just not auto-adding it. Should we do this, or let the error occur if the user explicitly "asked for it"?

Either way. I don't mind.

> I can test with different OTP versions, but not with different MySQL/MariaDB versions.

If we know that the code we change isn't DB specific, it's probably safe to test only one version.

Maybe it's not that hard to set up github CI with different versions. Someone just needs to look into it...

Maria-12648430 (approved):

Maria-12648430:
I have been running tests with OTP 21.3, 22.3 and 24.1 against MariaDB 10.5. They work, but I noticed there are no tests for using the socket backend at all yet. For my tests, I temporarily bastardized the one that uses local sockets, and I can confirm that it works with this PR, but not with master. So the changes made here are OK as far as I can tell, but a real test would be nice :)

juhlig:
I tested with MySQL 5.6 and 8.0, same OTP versions as @Maria-12648430. Looks fine there, too. Tests, ok, I'll see to that tomorrow :)

The GitHub CI seems to be tricky, especially MySQL... I may have a stab at it one of those days.

zuiderkwast:
OK awesome! Let's merge this one and we'll tackle CI later. If we just have one MariaDB CI job, it's better than nothing.

zuiderkwast (approved):
Test case looks good too. It's good to have. Merge whenever you want.

PR #185, opened 2022-01-04
fix: use ssl:versions as default ssl versions

suexcxine:
I encountered this error when I try to connect with ssl:
```
> ssl:connect(Port, [{versions,[tlsv1]},{verify,verify_none},{active,false}], 5000)
{error,{tls_alert,{bad_record_mac,"TLS client: In state hello at tls_record.erl:571 generated CLIENT ALERT: Fatal - Bad Record MAC\n {unsupported_version,{0,0}}"}}}
```
And the following is a similar PR from hackney:
https://github.com/benoitc/hackney/pull/619/files#diff-828780afc3d340b2952fd6442d1dd478aff3396ee750b236f5fe2a21145824ef

### DIFF
--- a/src/mysql_protocol.erl
+++ b/src/mysql_protocol.erl
@@ -421,15 +421,16 @@ maybe_do_ssl_upgrade(Host, gen_tcp, Socket0, SeqNum1, Handshake, SSLOpts,
                      Database, SetFoundRows) ->
     Response = build_handshake_response(Handshake, Database, SetFoundRows),
     {ok, SeqNum2} = send_packet(gen_tcp, Socket0, Response, SeqNum1),
-    case ssl_connect(Host, Socket0, SSLOpts, 5000) of
+    case ssl_connect(Handshake, Host, Socket0, SSLOpts, 5000) of
         {ok, SSLSocket} ->
             {ok, ssl, SSLSocket, SeqNum2};
         {error, Reason} ->
             exit({failed_to_upgrade_socket, Reason})
     end.
 
-ssl_connect(Host, Port, ConfigSSLOpts, Timeout) ->
-    DefaultSSLOpts0 = [{versions, [tlsv1]}, {verify, verify_peer}],
+ssl_connect(Handshake, Host, Port, ConfigSSLOpts, Timeout) ->
+    SSLVersions = determine_ssl_versions(Handshake),
+    DefaultSSLOpts0 = [{versions, SSLVersions}, {verify, verify_peer}],
     DefaultSSLOpts1 = case is_list(Host) andalso inet:parse_address(Host) of
         false -> DefaultSSLOpts0;
         {ok, _} -> DefaultSSLOpts0;
@@ -438,6 +439,17 @@ ssl_connect(Host, Port, ConfigSSLOpts, Timeout) ->
     MandatorySSLOpts = [{active, false}],
     MergedSSLOpts = merge_ssl_options(DefaultSSLOpts1, MandatorySSLOpts, ConfigSSLOpts),
     ssl:connect(Port, MergedSSLOpts, Timeout).
+
+
+%% @doc Determines which ssl versions to use according to server vendor and version
+%% since almostly, mysql < 5.7.0 and mariadb < 10.1.0 supports only tlsv1
+-spec determine_ssl_versions(#handshake{}) -> list().
+determine_ssl_versions(#handshake{server_vendor = mysql, server_version = Version}) when Version < [5, 7, 0] ->
+    [tlsv1];
+determine_ssl_versions(#handshake{server_vendor = mariadb, server_version = Version}) when Version < [10, 1, 0] ->
+    [tlsv1];
+determine_ssl_versions(_) ->
+    proplists:get_value(available, ssl:versions()).
 
 -spec merge_ssl_options(list(), list(), list()) -> list().
 merge_ssl_options(DefaultSSLOpts, MandatorySSLOpts, ConfigSSLOpts) ->


juhlig:
I think there was some fix in Ranch regarding ssl versions lately. I'll have a look there tomorrow to see if it might be related in any way and if there could be more to this.

zuiderkwast:
It looks like the default option `{versions, [tlsv1]}` was added when TLS support was first added in 8d322d48e4b76ce60633ab7e65e2c1daa0fb89a0, #69. We didn't discuss it in the comments. @fenek do you remember why this was added?

juhlig:
First, the Ranch fix is unrelated, it concerned some server options that were not allowed when TLSv1.3 was the only protocol.

<hr />

Here is what I found out in regards to TLS support in MySQL:
* Before 5.6.46, only TLSv1 was supported
* Up to 8.0.25, TLSv1, TLSv1.1 and TLSv1.2 were supported
* In 8.0.26, TLSv1 and TLSv1.1 were deprecated
* In 8.0.28, TLSv1 and TLSv1.1 support was removed

So I assume that @suexcxine (please confirm) either uses a 8.0.28 release of MySQL (or an older one where at least TLSv1 has been disabled via config). From looking at the `ssl` code in OTP, the error "`{unsupported_version,{0,0}}`" seems to indicate that there is no TLS version that both the client (only TLSv1 using our default option) and the server (only TLSv1.2) offer.

It further elicits why `tlsv1` has to be given as a default option. If no versions are given, `ssl` resorts to the content of the `supported` key returned from `ssl:versions/0`, which usually excludes the legacy versions (eg, there is only `'tlsv1.2'` if I check on my system). The `available` key instead lists _all_ versions that the current `ssl` implementation and backing crypto lib offer.

<hr />

Ok. Up to here, I can confirm that the change in the PR works, on a fairly recent system running MySQL 8.

But here is the catch: I dug out a very old VM that I had lying around, to try MySQL 5.6, and there it fails with the same error "`{unsupported_version,{0,0}}`". Specifically, this happens when `'tlsv1.2'` is in the versions list. I don't know if the problem is on the client or server side or in the crypto lib they both use (that's openssl 1.0.1f, it _should_ support TLSv1.2). It could also be possible that there is some kind of misconfiguration in the VM itself, after all it is _very_ old, running Ubuntu trusty.

However, if it is _not_ just a VM misconfiguration, we are facing a dilemma:
* On the one hand, we need to keep up with MySQL, so enabling TLSv1.2 is required. There is the workaround of giving `'tlsv1.2'` in the ssl options for that, but it's not nice.
* On the other hand, we are claiming support all the way back to MySQL 4.1 (or so the README says). If we apply the changes in this PR, things will suddenly break when used on older systems that use the default option (vs giving `tlsv1` explicitly).

@zuiderkwast, thoughts? =^^=

zuiderkwast:
Awesome findings Jan. Ideally we provide defaults depending on MySQL/MariaDB version, but we don't have this information in the TLS handshake of course. :-) :chicken: :egg: 

My first thought is that backward compatibility is more important than anything, but OTOH maybe nobody uses MySQL 5.6 anymore. I don't know. MariaDB might be more important. IIRC, Debian and other distros only ship MariaDB nowdays and many projects switched to it, so I think we should also consider how it behaves. It's also important that we have sensible defaults for the current versions.

We do still support MySQL 4.1 without TLS. Nobody used TLS in that time anyway. :-)

We should also provide info and links in our docs to the relevant webpages at MySQL and MariaDB so the users can be informed and set the ssl options manually.

That's my thoughts. @suexcxine and @juhlig do you have any ideas? Also involving @Maria-12648430.

juhlig:
> but we don't have this information in the TLS handshake of course

Wait, but we _do_! The _initial_ connection is made via plain TCP, and then _upgraded_ to SSL. So we do receive the initial handshake packet from the server, which includes the server version =D We just need to pass the handshake record on to `ssl_connect`, which is where those defaults are set.

> We do still support MySQL 4.1 without TLS. Nobody used TLS in that time anyway. :-)

Yeah, but I meant that if we claim compatibility back to 4.1, it would be awkward if we broke it for MySQL versions up to 5.6, ie waaay later ;)

zuiderkwast:
> The _initial_ connection is made via plain TCP, and then _upgraded_ to SSL. So we do receive the initial handshake packet from the server, which includes the server version =D We just need to pass the handshake record on to `ssl_connect`, which is where those defaults are set.

That's awesome!

I think the proper way to negotiate version is to provide all versions supported by the client, i.e. by OTP, so `{versions, proplists:get_value(available, ssl:versions())}`, but that is probably the same as not providing `versions` at all.

So, only for versions where MySQL (or MariaDB) somehow misbehaves (wrt TLS version negotiation), we'll need to pass `[tslv1]` or so.

suexcxine:
@juhlig this is my mysql version
```
MySQL [(none)]> select version();
+-----------+
| version() |
+-----------+
| 8.0.27    |
+-----------+
1 row in set (0.00 sec)

MySQL [(none)]> show variables like '%tls%';
+------------------------+-------------------------------+
| Variable_name          | Value                         |
+------------------------+-------------------------------+
| admin_tls_ciphersuites |                               |
| admin_tls_version      | TLSv1,TLSv1.1,TLSv1.2,TLSv1.3 |
| tls_ciphersuites       |                               |
| tls_version            | TLSv1,TLSv1.1,TLSv1.2,TLSv1.3 |
+------------------------+-------------------------------+
4 rows in set (0.00 sec)
```

suexcxine:
> I think the proper way to negotiate version is to provide all versions supported by the client, i.e. by OTP, so {versions, proplists:get_value(available, ssl:versions())}, but that is probably the same as not providing versions at all.
Here is the output of ssl:versions() on my system:
```
> ssl:versions().
[{ssl_app,"10.3.1.2"},
 {supported,['tlsv1.3','tlsv1.2']},
 {supported_dtls,['dtlsv1.2']},
 {available,['tlsv1.3','tlsv1.2','tlsv1.1',tlsv1]},
 {available_dtls,['dtlsv1.2',dtlsv1]},
 {implemented,['tlsv1.3','tlsv1.2','tlsv1.1',tlsv1]},
 {implemented_dtls,['dtlsv1.2',dtlsv1]}]
```
not providing versions at all means ['tlsv1.3','tlsv1.2'] and `proplists:get_value(available, ssl:versions()` means ['tlsv1.3','tlsv1.2','tlsv1.1',tlsv1], which is more permissive.

And with regard to compatibility issue, IMHO, since at least we could use manual version options, we could say we are compatible even with MySQL 4.1 to 5.6.

zuiderkwast:
For MariaDB, it looks like you can configure which TLS versions the server supports. It also depends on which TLS library is used (OpenSSL and others) https://mariadb.com/kb/en/secure-connections-overview/

```
[mariadb]
...
tls_version = TLSv1.2,TLSv1.3
```

juhlig:
> it's 8.0.27, not 8.0.28, and it looks like TLSv1 is not disabled

That's strange then... the MySQL documentation [here](https://dev.mysql.com/doc/refman/8.0/en/encrypted-connection-protocols-ciphers.html) says that TLS versions could further be restricted at the host level:

> For example, your system may have an /etc/ssl/openssl.cnf file that contains these lines to restrict TLS protocols to TLSv1.2 or higher:
> ```
> [system_default_sect]
> MinProtocol = TLSv1.2
> ```

Can you check that, please?

juhlig:
> For MariaDB, it looks like you can configure which TLS versions the server supports. It also depends on which TLS library is used (OpenSSL and others) https://mariadb.com/kb/en/secure-connections-overview/
> 
> ```
> [mariadb]
> ...
> tls_version = TLSv1.2,TLSv1.3
> ```

Yes, I just found that, too. MySQL is more or less the same: https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_tls_version

There is another problem with the server version returned by the initial handshake packet: it does not say if it is MariaDB or MySQL. The MariaDB docs say it should by default be prefixed by `5.5.5-` if it is MariaDB, not sure what "by default" means here :\

juhlig:
> The MariaDB docs say it should by default be prefixed by `5.5.5-` if it is MariaDB, not sure what "by default" means here :\

Ok, so that is a hack XD Up to 5.5, MariaDB versions followed MySQL versions, with MariaDB 10 the 5.5.5- prefix was added because otherwise the replication protocol would break. This means we also have to fix the version number parsing, but we can also distinguish MySQL from MariaDB this way.



> If it's configurable, it means we can't fully rely on the MySQL version we get from the initial handshake.

I think we can. You probably can't configure a TLS version lower or higher than what is supported by MySQL/MariaDB. So for MySQL pre-5.6.46, we use TLSv1 only, for versions up to 8.0.15 we use TLSv1-1.2, and for everything greater we use all that we have available (that may work for everything >=5.6.46 even). We can't be sure about MariaDB, though, the docs don't go before 10.0.15 in that regard, and it looks like it only depends on the crypto lib.

suexcxine:
@juhlig  you are right
```
[ssl_sect]
system_default = system_default_sect

[system_default_sect]
MinProtocol = TLSv1.2
CipherString = DEFAULT@SECLEVEL=2
```
but we are only using the mysql:8.0.27 image from docker hub
and I don't think they tweaked this openssl configuration,
so it's probably setting MinProtocol = TLSv1.2 as default

suexcxine:
What if we run MySQL 5.6 in a recent os such as ubuntu 20.04 or debian buster ? on which openssl's default configuration is MinProtocol=TLSv1.2. 

if tlsv1 doesn't work on such system, IMHO we don't need to care about because anyway we can't ensure full compatibility using only default ssl options.

juhlig:
FWIW: https://mariadb.org/state-ssl-mariadb/. So MariaDB supports 1.1 and 1.2 beginning with .5.41 in the 5 series, and .0.15 in the 10 series.

I just put up #187 which detects the server vendor from the version (except two cases, 5.1.x and 5.5.x; I simply assume it to be MySQL then, which seems more likely?)

I suggest we do the following: If vendor is MySQL and version is <5.6.46, or vendor is MariaDB and version is <10.0.15, we default to only TLSv1. Otherwise, we use all that is available as the default. There is a gap for MariaDB, 5.5.41+ where we _could_ use all available, but as I said, in 5.5 we can't really decide which is which.

What do you think?

juhlig:
> What if we run MySQL 5.6 in a recent os such as ubuntu 20.04 or debian buster ? on which openssl's default configuration is MinProtocol=TLSv1.2.

I don't know if anyone would want to do that, but anyway, then we can't do anything about it. We do SSL upgrade if...
* the user gave SSL options, and
* the server says it has SSL enabled via the CLIENT_SSL capability flag (I just noticed that actually we don't... we ignore that and try SSL anyway; that's a bug I'll address tomorrow)

zuiderkwast:
> I suggest we do the following: If vendor is MySQL and version is <5.6.46, or vendor is MariaDB and version is <10.0.15, we default to only TLSv1. Otherwise, we use all that is available as the default. There is a gap for MariaDB, 5.5.41+ where we *could* use all available, but as I said, in 5.5 we can't really decide which is which.

If a server only supports TLS 1.0 and we send a version list like `['tlsv1.3','tlsv1.2','tlsv1.1',tlsv1]`, the version negotiation *should work* and 1.0 should be selected. Only in server versions which don't negotiate correctly, we need to set some other default. It would be interesting to find out which db versions can negotiate correctly and which can't, so that we can manipulate the version list only when necessary. (It would be very nice to have a CI with multiple DB versions now.)

There seem to be some differences in how the negotiation actually works depending on RFC. The TLS 1.3 RFC https://datatracker.ietf.org/doc/html/rfc8446 even includes some changes to TLS 1.2 so there is some mess here.

Even if `all_available` works for the most recent db versions, perhaps it's not a perfect default. There is a reason OpenSSL and OTP have updated the defaults to only include TLS 1.2 and TLS 1.3. They are pushing clients and servers to phase out the older versions.

zuiderkwast:
Let's not care too much about the versions which already passed end-of-life ([Maria](https://en.wikipedia.org/wiki/MariaDB#Versioning) 10.2 and below, [My](https://en.wikipedia.org/wiki/MySQL#Release_history) 5.6 and below). We can prioritize the recent versions I think.

juhlig:
@suexcxine @zuiderkwast  I see that #187 was merged into master. So can you please rebase your PR on current master, and set default protocols according to vendor and version? I think `[tlsv1]` for MySQL <5.7.0 and MariaDB <10.1.0, and `proplists:get_value(available, ssl:versions())` for everything else?

suexcxine:
ok, I'll do this tonight.

suexcxine:
@juhlig done, please have a look.

juhlig:
> @juhlig done, please have a look.

Looks good to me :)

juhlig (approved):

zuiderkwast:
The updated code looks fine. I'm just not 100% convinced that it solves the problem in all cases.

What about DB versions like MySQL 5.7 or 8.0.0 which can handle for example 1, 1.1 and 1.2? Are we sure they're able to negotiate a version if we send `[tslv1, 'tlsv1.1', 'tlsv1.2', 'tlsv1.3']`?

@suexcxine It looks like your DB actually supports TLSv1 so when we send `[tlsv1]` (without this PR) I still don't understand why it's unable to negotiate and select TLSV1...

In MariaDB (recent versions) you can use different TLS implementations and they don't support the same set of TLS versions. I just hope they're able to negotiate properly.

suexcxine:
my case is this:
```
[ssl_sect]
system_default = system_default_sect

[system_default_sect]
MinProtocol = TLSv1.2
CipherString = DEFAULT@SECLEVEL=2
```
`MinProtocol = TLSv1.2` set in /etc/ssl/openssl.cnf

we are using a mysql:8.0.27 image from docker hub
and it's based on debian buster,
and when you do `apt-get install openssl` (1.1.1d) in debian buster
the default config is MinProtocol = TLSv1.2

that's why `[tlsv1]` doesn't work.

zuiderkwast:
> `MinProtocol = TLSv1.2` set in /etc/ssl/openssl.cnf

@suexcxine OK, good, this explains your problem.

> If we apply the changes in this PR, things will suddenly break when used on older systems that use the default option (vs giving tlsv1 explicitly).

@juhlig How could it break? Do we know for sure that old MySQL is confused if we always send all available?

juhlig:
> @juhlig How could it break? Do we know for sure that old MySQL is confused if we always send all available?

I don't know, but it does. I set up a fresh Ubuntu 14 VM with MySQL 5.6.33 because I suspected a misconfiguration in the one I had tested before, but it's the same there.

> [Edit] As long as `tlsv1` is among the versions we send, I'd be very surprised if anything stops working.

In the test with the new old VM, it stopped working as soon as `tlsv1.2` was among the versions. I haven't tested with MySQL 5.7, though, so I don't know how it behaves.

> All I'm saying is that we might want to avoid cluttering the code with such clauses, given those DB versions are not maintained anymore and users can always avoid upgrading the client or provide a TLS versions config explicitly.

I'm fine with that if you are ;)

zuiderkwast:
> it stopped working as soon as tlsv1.2 was among the versions.

OK, this convinces me that we need this logic to support 5.6.

> I haven't tested with MySQL 5.7, though, so I don't know how it behaves.

I think we can merge this PR now, but I would really like to have some CI jobs to test this on 5.7 and other versions before we release a new version........

PR #190, opened 2022-01-26
fix: function clause error in nesting transactions

suexcxine:
We encountered following error:
```
Uncaught exit - {{:function_clause, [{:mysql_conn, :handle_call, [:commit, {#PID<0.18851.2939>, #Reference<0.517344270.338165767.72012>}, {:state, [8, 0, 13], 2757790, #Port<0.23126779>, :gen_tcp, [], :undefined, 'example.com', 3306, "meiqia_ticket", "yourpassword", "tickets", [], [], "mysql_native_password", <<123, 8, 94, 45, 28, 42, 28, 94, 125, 65, 123, 23, 108, 101, 116, 48, 125, 60, 109, 67, 0>>, [], true, false, 5000, :infinity, :infinity, 60000, 0, 3, 0, 0, [{#PID<0.23837.2554>, #Reference<0.517344270.1424752644.79848>}, {#PID<0.18851.2939>, #Reference<0.517344270.338165767.71948>}], :undefined, {:dict, 0, 16, 16, 8, 80, 48, {[], [], ...}, {{...}}}, {:cache, {1642, 754754, 337352}, {:dict, 3, 16, 16, 8, 80, ...}}, false, false}], [file: 'src/mysql_conn.erl', line: 294]}, {:gen_server, :try_handle_call, 4, [file: 'gen_server.erl', line: 715]}, {:gen_server, :handle_msg, 6, [file: 'gen_server.erl', line: 744]}, {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 226]}]}, {:gen_server, :call, [#PID<0.2428.0>, :commit, :infinity]}}
```
This happens when transaction a and then transaction b are started sequentially on a connection,
and transaction a commits/rollbacks before transaction b.

This could be reproduced by following snippet
```
{:ok, pid} = :mysql.start_link([{:host,'127.0.0.1'},{:user,<<"test">>},{:password,<<>>},{:database,<<"test">>}])
f1 = fn ->
  Process.sleep(500)
  :mysql.query(pid, "show databases;")
end

f2 = fn ->
  Process.sleep(1000)
  :mysql.query(pid, "show databases;")
end
spawn(fn -> {:atomic, _} = :mysql.transaction(pid, f2) end); {:atomic, _} = :mysql.transaction(pid, f1)
```

### DIFF
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -68,7 +68,7 @@
 -type query_result() :: ok
                       | {ok, [column_name()], [row()]}
                       | {ok, [{[column_name()], [row()]}, ...]}
-                      | {error, server_reason()}.
+                      | {error, server_reason() | busy}.
 
 -type transaction_result(Result) :: {atomic, Result} | {aborted, Reason :: term()}.
 
@@ -366,6 +366,9 @@ query(Conn, Query, Params, FilterMap) when (Params == no_params orelse
 %% For queries that don't return any rows (INSERT, UPDATE, etc.) only the atom
 %% `ok' is returned.
 %%
+%% If this function is called on a connection which is already in transaction 
+%% owned by another process, `{error, busy}` will be returned.
+%%
 %% === FilterMap details ===
 %%
 %% If the `FilterMap' argument is used, it must be a function of arity 1 or 2
@@ -498,6 +501,9 @@ execute(Conn, StatementRef, Params, FilterMap) when FilterMap == no_filtermap_fu
 %%
 %% See `query/5' for an explanation of the `FilterMap' argument.
 %%
+%% Note that if this function is called on a connection which is already in transaction 
+%% owned by another process, `{error, busy}` will be returned.
+%%
 %% @see prepare/2
 %% @see prepare/3
 %% @see prepare/4
@@ -618,6 +624,10 @@ transaction(Conn, Fun, Retries) ->
 %% can be nested and are restarted automatically when deadlocks are detected.
 %% MySQL's savepoints are used to implement nested transactions.
 %%
+%% If this function is called on a connection which is already in transaction,
+%% `{aborted, busy}` will be returned. Since the idea of "nested transaction" is
+%% that this function can be called in the Fun, but all within the same process.
+%%
 %% Fun must be a function and Args must be a list of the same length as the
 %% arity of Fun.
 %%
@@ -665,8 +675,12 @@ transaction(Conn, Fun, Args, Retries) when is_list(Args),
                                            is_function(Fun, length(Args)) ->
     %% The guard makes sure that we can apply Fun to Args. Any error we catch
     %% in the try-catch are actual errors that occurred in Fun.
-    ok = gen_server:call(Conn, start_transaction, infinity),
-    execute_transaction(Conn, Fun, Args, Retries).
+    case gen_server:call(Conn, start_transaction, infinity) of
+        ok -> 
+            execute_transaction(Conn, Fun, Args, Retries);
+        {error, busy} ->
+            {aborted, busy}
+    end.        
 
 %% @private
 %% @doc This is a helper for transaction/2,3,4. It performs everything except
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -624,9 +624,10 @@ transaction(Conn, Fun, Retries) ->
 %% can be nested and are restarted automatically when deadlocks are detected.
 %% MySQL's savepoints are used to implement nested transactions.
 %%
-%% If this function is called on a connection which is already in transaction,
-%% `{aborted, busy}` will be returned. Since the idea of "nested transaction" is
-%% that this function can be called in the Fun, but all within the same process.
+%% If this function is called on a connection which is already in a transaction
+%% owned by another process, `{aborted, busy}` is returned. The idea of nested
+%% transactions is that this function can be called in the Fun, but all within
+%% the same process.
 %%
 %% Fun must be a function and Args must be a list of the same length as the
 %% arity of Fun.
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -302,6 +302,11 @@ handle_call(Msg, From, #state{socket = undefined} = State) ->
         {error, _} = E ->
             {stop, E, State}
     end;
+handle_call({query, _Query, _FilterMap, _Timeout}, {FromPid, _},
+            State = #state{transaction_levels = [{OtherFromPid, _} | _]})
+  when FromPid =/= OtherFromPid ->
+    %% this conn is currently in transaction owned by another process
+    {reply, {error, busy}, State};
 handle_call({query, Query, FilterMap, Timeout}, _From, State) ->
     {Reply, State1} = query(Query, FilterMap, Timeout, State),
     {reply, Reply, State1};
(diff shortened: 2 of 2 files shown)


juhlig:
I may be mistaken, but I'm pretty sure that this (running transactions in parallel on the same connection) is not something that should be done at all, and we should rather do something to _forbid_ it instead of making it somehow (seem to) work.

The actual idea of "nested transactions" is that `mysql:transaction` can be called in the transaction function, but all within the same process

If you use `transaction` from several processes in parallel, well, it has "race condition" written all over it. Depending on timing, the two transactions A and B might be executed one after the other, but it might be A then B or B then A. Or they may be mixing into each other, B might be executed nested within A or A nested within B, and the statements within the transaction functions might also be executed in a variety of orders. In one word: mayhem!

zuiderkwast:
It's modelled after mnesia's transactions. How does mnesia behave in this scenario?

suexcxine:
I agree with @juhlig . In fact, what I did in my product to avoid this issue is to make a mysql conn unseen by other clients before its transaction commits.

suexcxine:
So we should raise an exception immediately when a second process tries to start transaction on a mysql connection whose status is already in transaction.
Am I right ?

zuiderkwast:
> So we should raise an exception immediately when a second process tries to start transaction on a mysql connection whose status is already in transaction.
> Am I right ?

@suexcxine I agree.

If we make mysql:transaction/N return an error in caller's process without killing the connection process, it would be nice. Then, we don't have to kill the mysql_conn process and the other process who owns the transaction.

We have the pid stored in the state in mysql_conn:

```Erlang
handle_call(start_transaction, {FromPid, _}, ...) ->
    ...
    {reply, ok, State1#state{transaction_levels = [{FromPid, MRef} | L]}};
```

I think we can use that. We already pattern-match on the FromPid in the rollback and commit clauses.

Btw, instead of a list `transaction_levels = [{FromPid, MRef}]`, it should be enough to store one pid and one monitor, i.e. we only need to store a tuple `transaction_state = {Pid, MRef, N}` where N is the number of levels. @juhlig WDYT?

Regarding how we can make `mysql:transaction/4` return an error to the caller without killing the processes: `handle_call(start_transaction, ...)` can reply with an error or `busy` (not `ok`). Then `mysql:transaction/4` can return something like `{aborted, busy}`. Makes sense?

src/mysql_conn.erl
```diff
@@ -477,13 +478,26 @@ handle_call(rollback, {FromPid, _},
                                                ?cmd_timeout),
     setopts(SockMod, Socket, [{active, once}]),
     State1 = update_state(Res, State),
-    {reply, ok, State1#state{transaction_levels = L}};
+    {reply, ok, State1#state{transaction_levels = L1}};
+handle_call(rollback, {FromPid, _},
+            State = #state{status = Status, transaction_levels = TransLevels})
+  when Status band ?SERVER_STATUS_IN_TRANS /= 0 ->
+    % This case happens when we start transaction a and then transaction b
```
zuiderkwast, line 485: Just a style nit: Use `%%` for comments aligned with the rest of the code. Use `%` for end-of-line comments. This is by the [Emacs mode](https://www.erlang.org/doc/man/erlang.el.html#indent) (which is the official Erlang style):

> * Lines with one %-character is indented to the right of the code. The column is specified by the variable comment-column, by default column 48 is used.
> * Lines with two %-characters will be indented to the same depth as code would have been in the same situation.
> * Lines with three of more %-characters are indented to the left margin.

Maria-12648430:
> > So we should raise an exception immediately when a second process tries to start transaction on a mysql connection whose status is already in transaction.
> > Am I right ?
> 
> @suexcxine I agree.

Um, guys... I think you're missing a point here :sweat_smile:

Forbidding concurrent transactions is not enough. Even if only one process (A) is running a transaction, if another process (B) issues a simple query via `mysql:query` (ie not as part of a `mysql:transaction`), that query will sneak into the transaction, meaning it will not be committed until A commits the transaction, and it may even be rolled back, but B will never know that, from his point of view the query was executed the moment he received the `ok`.

So, _any_ action like `query` or `execute` etc must be rejected when the connection is in a transaction, if it comes from a different process than the one running the transaction.

Do you agree?

juhlig:
@zuiderkwast 

> Btw, instead of a list `transaction_levels = [{FromPid, MRef}]`, it should be enough to store one pid and one monitor, i.e. we only need to store a tuple `transaction_state = {Pid, MRef, N}` where N is the number of levels. @juhlig WDYT?

Yes, I agree. With that, it might even be possible (don't know about useful) for a process to hand off (`give_away`?) a connection-in-transaction to another process simple by changing the Pid and Monitor.

<hr />

@Maria-12648430 aw, damn... guess you're right =(

zuiderkwast:
> So, _any_ action like `query` or `execute` etc must be rejected when the connection is in a transaction, if it comes from a different process than the one running the transaction.

@Maria-12648430 I agree. The worst case is queries like `ALTER TABLE` which can even cause an ongoing transaction to be implicitly committed. Query and execute could return `{error, busy}` in this case.

> it might even be possible (don't know about useful) for a process to hand off (`give_away`?) a connection-in-transaction to another process simply by changing the Pid and Monitor.

@juhlig `give_away` is naturally a good name for this, but I don't think it's useful. :-)

juhlig:
> Query and execute could return `{error, busy}` in this case.

I would prefer something like `in_transaction`, that would be clearer I guess.

> but I don't think it's useful. :-)

Yeah, probably not ;)

zuiderkwast:
@juhlig  We have a function called `in_transaction/1`, so `{error, in_transaction}` can also be confusing. More clear but too long: `{error, in_transaction_owned_by_another_process}`. I think I prefer `busy` for now and that we document what it means.

An alternative is to automatically wait for the other process' transaction to finish. It would involve a queue of waiting commands to be stored in the conn process' state. When a transaction is finished, these messages can be handled.

suexcxine:
I submitted a new version, please have a look.

zuiderkwast (commented):
Great! I think `{error, busy}` should be explained at `execute/5` too, or refer to query/N for an explanation of all return values or something like that.

suexcxine:
Thanks for the review! code updated.

src/mysql.erl
```diff
@@ -618,6 +624,10 @@ transaction(Conn, Fun, Retries) ->
 %% can be nested and are restarted automatically when deadlocks are detected.
 %% MySQL's savepoints are used to implement nested transactions.
 %%
+%% If this function is called on a connection which is already in transaction,
+%% `{aborted, busy}` will be returned. Since the idea of "nested transaction" is
+%% that this function can be called in the Fun, but all within the same process.
```
zuiderkwast, line 629: "owned by another process" is missing here.

```suggestion
%% If this function is called on a connection which is already in transaction
%% owned by another process, `{aborted, busy}` will be returned. The idea of
%% "nested transaction" is that this function can be called in the Fun, but all
%% within the same process.
```

zuiderkwast (commented):
:+1:

juhlig:
Looks good to me :+1:

suexcxine:
suggestion applied :)

zuiderkwast:
Great! I guess this is ready to squash-merge then!

zuiderkwast:
Updated title and top comment for traceability. I also edited the commit message.

PR #191, opened 2022-02-04
Obfuscate password and critical SSL options

juhlig:
When a `mysql_conn` process crashes, sensitive information can end up in logs.

This implementation of `format_status` replaces the value of `password` in the state record with the atom `hidden`, and likewise replaces the values for keys `key`, `cert` and `cacerts` with `hidden` in the SSL options.

The obfuscation only takes place if the process terminates (abnormally) which is where logging typically appears, ie the values are not touched when `sys:get_status` is used. The "real" state can be retrieved when `sys:get_state` is used, anyway.

### DIFF
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -28,7 +28,7 @@
 
 -behaviour(gen_server).
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2,
-         code_change/3]).
+         code_change/3, format_status/2]).
 
 -define(default_host, "localhost").
 -define(default_port, 3306).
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
Good idea! I haven't seen `format_status` before but it seems to have existed since OTP 13.

Is it worth hiding username too?

src/mysql_conn.erl
```diff
@@ -830,3 +830,19 @@ demonitor_processes(List, 0) ->
...
+format_status(normal, [_PDict, State]) ->
+	{data, [{"State", State}]};
+format_status(terminate, [_PDict, State=#state{ssl_opts=undefined}]) ->
+	{data, [{"State", State#state{password = hidden}}]};
+format_status(terminate, [_PDict, State=#state{ssl_opts=SSLOpts}]) ->
+	SSLOpts1 = lists:map(
+		fun
+			({cert, _}) -> {cert, hidden};
+			({key, _}) -> {key, hidden};
+			({cacerts, _}) -> {cacerts, hidden};
```
zuiderkwast, line 843: Certificates only contain public keys and other non-secret stuff. Only `key` is secret and needs to be hidden IUCC.

juhlig:
> Is it worth hiding username too?

Maybe, not sure. Probably won't hurt to hide it ;)

src/mysql_conn.erl
juhlig: I just looked at what Ranch hides, TBH ;)

zuiderkwast:
You can decide, as long as we hide password and TLS key. Username and certificates are less secret. I don't mind if we hide them too but I don't mind if we don't either.

juhlig:
In that case, let's leave it as it is? The username may be necessary for a user to figure out which of his possibly many connections has a problem. OTOH, cert and cacerts don't provide anything useful but will just create noise in logs.

zuiderkwast (approved):

PR #199, opened 2023-03-01
fix {badmatch,{error,closed}},[{mysql_protocol,prepare,3

leeyisoft:
2023-03-01 00:15:27.479 UTC [error] <0.5808.12> gen_server user_server terminated with reason: {{{badmatch,{error,closed}},[{mysql_protocol,prepare,3,[{file,"src/mysql_protocol.erl"},{line,215}]},{mysql_conn,handle_call,3,[{file,"src/mysql_conn.erl"},{line,319}]},{gen_server,try_handle_call,4,[{file,"gen_server.erl"},{line,721}]},{gen_server,handle_msg,6,[{file,"gen_server.erl"},{line,750}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,226}]}]},{gen_server,call,[<0.5806.12>,{param_query,<<"SELECT `device_name` FROM `user_device`\n        WHERE `status` = ? AND `user_id` = ? A...">>,...},...]}} in gen_server:call/3 line 247

### DIFF
--- a/src/mysql_protocol.erl
+++ b/src/mysql_protocol.erl
@@ -212,7 +212,16 @@ fetch_query_response(SockModule, Socket, AllowedPaths, FilterMap, Timeout) ->
 prepare(Query, SockModule, Socket) ->
     Req = <<?COM_STMT_PREPARE, (iolist_to_binary(Query))/binary>>,
     {ok, SeqNum1} = send_packet(SockModule, Socket, Req, 0),
-    {ok, Resp, SeqNum2} = recv_packet(SockModule, Socket, SeqNum1),
+    case recv_packet(SockModule, Socket, SeqNum1) of
+        {error, #error{code = Code, state = State, msg = Msg}} ->
+            #error{code = Code, state = State, msg = Msg};
+        {error, Error} ->
+            #error{code = -1, state = #{}, msg = Error};
+        {ok, Resp, SeqNum2} ->
+            prepare_recv_packet(Resp, SeqNum2, Query, SockModule, Socket)
+    end.
+
+prepare_recv_packet(Resp, SeqNum2, Query, SockModule, Socket) ->
     case Resp of
         ?error_pattern ->
             parse_error_packet(Resp);
@@ -1277,9 +1286,9 @@ recv_packet(SockModule, Socket, Timeout, ExpectSeqNum, Acc) ->
                 SeqNum :: integer()) ->
     {ok | {error, Reason}, NextSeqNum :: integer()}
     when Reason :: not_allowed
-	         | file:posix()
-		 | badarg
-		 | system_limit.
+             | file:posix()
+         | badarg
+         | system_limit.
 send_file(SockModule, Socket, Filename, AllowedPaths, SeqNum0) ->
     {Result, SeqNum1} = case allowed_path(Filename, AllowedPaths) andalso
                              file:open(Filename, [read, raw, binary]) of


zuiderkwast (changes requested):
Hi!

What happens after this "fix"? The connection is down, but the process appears to be up? What do you expect the user to do after that?

If you want a better error in the log, we should move the exit to mysql_conn, like i suggested in this PR #173.

We should do the same for query, prepare and execute.

leeyisoft:
I agree with that: We should do the same for query, prepare and execute.

I want MySQL clients to be able to sense that "MySQL server terminates tcp" so that "business applications" can re-apply for a tcp and complete MySQL transactions

My application scenario is as follows: a supervisor monitors several gen_servers as workers. When the worker receives messages (in case of a small number of messages, it will not receive messages for more than 2 hours), the worker interacts with MySQL. After the MySQL server terminates tcp, a message is sent. In this case, {error,closed} appears. I want to make my worker aware of {error,closed} without killing the worker (in this case, when the worker dies, the message will be lost, so MQ will not be used for the time being).

leeyisoft:
Thank you for giving me a try catch demo. I will add a try catch to my code and observe it for a while.

I'm sure it's very necessary that "we would need to return the error from mysql_protocol to mysql_conn and there we do  exit(Reason).", but I'm not familiar with mysql_protocol, and erlang is new to it, so merge it if you find it useful, and close it if it doesn't

leeyisoft:
Excuse me, how does my pattern match the case of MySQL server shutting down tcp

```
-spec execute_try(Sql :: any()) ->
          {ok, LastInsertId :: integer()} | {error, any()}.
execute_try(Sql) ->
    try
        poolboy:transaction(mysql, fun(Pid) ->
               case mysql:query(Pid, Sql) of
                   ok ->
                       {ok, mysql:insert_id(Pid)};
                   Res ->
                       Res
               end
        end)
    catch
        %  Excuse me, I don't know how the patterns match up here
        Class1:Reason1:Stacktrace1 ->
            execute_try(Sql);
        Class:Reason:Stacktrace ->
            erlang:raise(Class, Reason, Stacktrace)
    end.
```

zuiderkwast:
The exit reason is included in your first message :)

> 2023-03-01 00:15:27.479 UTC [error] <0.5808.12> gen_server user_server terminated with reason: {{{badmatch,{error,closed}},[{mysql_protocol,prepare,3,[{file,"src/mysql_protocol.erl"},{line,215}]},
> ...

But you don't need to check all of it. This should be enough:

```Erlang
try
    ....
catch
    exit:{{{badmatch, {error, closed}}, _}, _} ->
        execute_try(Sql)
end.
```

No need to catch the stacktrace.

Also don't catch and raise the exceptions you're not interested in. Not catching them has the same effect.

Btw, I think you want to test your code. You can make MySQL close a connection if you open a `mysql` connection in the terminal and type **SHOW PROCESSLIST;** to find the erlang connection and then **KILL *connection id*;**. You can make poolboy have only one connection to make it easier to find it. Or kill all :).

leeyisoft:
Thanks for your detailed introduction, and how to test the code quickly.

Seeing this sentence, I was shocked:
“Also don't catch and raise the exceptions you're not interested in.  Not catching them has the same effect.”

zuiderkwast:
OK, great.

> Seeing this sentence, I was shocked:
> “Also don't catch and raise the exceptions you're not interested in. Not catching them has the same effect.”

Really? There is no point to catch something and not do anything with it and just throw it again. The following two have exactly the same effect.

* `try foo() catch C:E:S -> erlang:raise(C, E, S) end`
* `foo()`

PR #204, opened 2023-07-10
Change test passwords to pass STRONG validation policy

juhlig:
As discussed in #203. I changed the `.travis.yml` file as well, even though we don't use Travis any longer, for good measure.


zuiderkwast (approved):
LGTM

PR #206, opened 2024-10-07
Allow binary as the name of prepared statement

zmstone:
This is to avoid creating atoms when making use of prepared statements.

### DIFF
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -385,10 +385,13 @@ handle_call({prepare, Query}, _From, State) ->
             State2 = State#state{stmts = Stmts1},
             {reply, {ok, Id}, State2}
     end;
-handle_call({prepare, Name, Query}, _From, State) ->
+handle_call({prepare, Name, Query}, _From, State) when is_atom(Name);
+                                                       is_binary(Name) ->
     {Reply, State1} = named_prepare(Name, Query, State),
     {reply, Reply, State1};
-handle_call({unprepare, Stmt}, _From, State) ->
+handle_call({unprepare, Stmt}, _From, State) when is_atom(Stmt);
+                                                  is_integer(Stmt);
+                                                  is_binary(Stmt) ->
     case dict:find(Stmt, State#state.stmts) of
         {ok, StmtRec} ->
             #state{socket = Socket, sockmod = SockMod} = State,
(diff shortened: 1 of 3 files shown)


src/mysql_conn.erl
```diff
@@ -385,11 +385,10 @@ handle_call({prepare, Query}, _From, State) ->
...
             {reply, {ok, Id}, State2}
     end;
-handle_call({prepare, Name, Query}, _From, State) when is_atom(Name) ->
+handle_call({prepare, Name, Query}, _From, State) ->
     {Reply, State1} = named_prepare(Name, Query, State),
     {reply, Reply, State1};
-handle_call({unprepare, Stmt}, _From, State) when is_atom(Stmt);
-                                                  is_integer(Stmt) ->
+handle_call({unprepare, Stmt}, _From, State) ->
     case dict:find(Stmt, State#state.stmts) of
```
zuiderkwast, line 392: I don't remember this, but do we use an integer for unnamed prepared statements, i.e. query with parameters that implicitly become a prepared statement? What happens if a user uses an integer as the name of a prepared statement? Is there a risk of collission?

If yes, then maybe we should at least forbid integers as the name, like `when not is_integer(Name)`?

zuiderkwast (commented):
Hi @zmstone.

You're right, it's not good to create a lot of atoms. But it's also not good to create a lot of prepared statements. Do you generate statements automatically and risk generating a lot of them?

zmstone:
> Hi @zmstone.
> 
> You're right, it's not good to create a lot of atoms. But it's also not good to create a lot of prepared statements. Do you generate statements automatically and risk generating a lot of them?

I do not see a risk of generating a lot of them in a short run, but more of a risk in a long term.
In our a long running system (EMQX), users can create/update/delete data integrations with MySQL.
We also generate random IDs for health-check probing of the user inputs.

src/mysql_conn.erl
zmstone: I can add and extra `is_binary(Stmt)` to be on the safe side.

zmstone:
@zuiderkwast thank you for the quick review.

We have accumulated quite a lot of changes in our fork here: https://github.com/emqx/mysql-otp
I'll try to break them down in to smaller change sets and send more PRs upstream.
My goal is to minimize the divergence in our fork.

src/mysql_conn.erl
zmstone: ah, you meant not to allow user explicitly unprepare using an integer.
that's new to me, I don't know why this clause was added here in the first place.

zuiderkwast (approved):

PR #207, opened 2024-10-07
GitHub Actions

zmstone:
Changes in this PR:

- Add GitHub Actions for continuous integration.
  Improve local test experience by using docker to start MySQL or MariaDB.
  Added scripts to bootstrap test environment in scripts dir. 
- Changed `0.0` matches to `+0.0` to make it compile on OTP 27.
- Made decimal test case to work with MySQL >= 8.

Will send a follow-up PR to delete old CI files.
GitHub Actions in this repo is only triggered after this PR is merged, to review the CI runs: https://github.com/zmstone/mysql-otp/actions


juhlig:
I'm on vacation this week and can't do a conclusive review before I'm back 🫤

zmstone:
https://github.com/mysql-otp/mysql-otp/pull/207/commits/7e53252cbc191149af5dcfae9f1e8249a63eb153

I am buzzed by this test case. Not sure how it might have worked in the first place.
Need some pointers.

.ci/prep.sh
```diff
@@ -0,0 +1,43 @@
...
+export MYSQL_VERSION="${MYSQL_VERSION:-8.4}"
+export MYSQL_CERTS_DIR='/etc/mysql_certs'
+
+mkdir -p .ci/run .ci/certs
+SSLDIR=/etc/mysql_certs make tests-prep
+sudo cp test/ssl/ca.pem .ci/certs/
+sudo mv test/ssl/server-key.pem .ci/certs/
+sudo mv test/ssl/server-cert.pem .ci/certs/
+mv test/ssl/my-ssl.cnf .ci/
+sudo chmod 660 .ci/certs/*
```
zuiderkwast, line 15: This is for running the tests using docker, not necessarily in a CI but also locally?

Why does this need sudo? Do the certs need to be owned by root? sudo may prompt for password so maybe it can be a problem to use it in the script? (In Github Actions, it's configured to not need a password I guess, but if we want to use this from some other place, such as running it locally.)

zuiderkwast (commented):
Very nice!

I don't know what's the problem with the test case for decimals that was commented out. I don't remember any problem with it in the past. It was a long time ago.

.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+    branches:
+      - master
+jobs:
+  ci:
+    name: CI
+    runs-on: ${{matrix.os}}
+    strategy:
+      fail-fast: false
+      matrix:
+        otp: [25, 26]
```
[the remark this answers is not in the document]
zmstone: added 24

.ci/prep.sh
zmstone: added a comment to explain why sudo is needed here

zmstone:
> I don't know what's the problem with the test case for decimals that was commented out. I don't remember any problem with it in the past. It was a long time ago.

@zuiderkwast  thank you for the review.
I need a pointer to see how this test worked before, and compare the test runs to see why it's not working now.
I tried to test with MySQL 8.4, 8.0 and 5.7, none of them worked.

zmstone:
> > I don't know what's the problem with the test case for decimals that was commented out. I don't remember any problem with it in the past. It was a long time ago.
> 
> @zuiderkwast thank you for the review. I need a pointer to see how this test worked before, and compare the test runs to see why it's not working now. I tried to test with MySQL 8.4, 8.0 and 5.7, none of them worked.

Actually, re-ran it against 5.7, it worked there.
On 8.0:
```
**error:{badmatch,{ok,ok,
              [{warning_msg,"Note 1265: Data truncated for column 'balance' at row 1\n in UPDATE test_decimals SET balance = balance - ? WHERE id = ?\n"}]}}
  output:<<"">>
```

README.md
```diff
@@ -117,33 +117,7 @@ To run individual test suites, use `make eunit t=SUITE` where SUITE is one of
...
-```
-
-Before running the test suite `ssl_tests` you'll also need to generate SSL files
-and MySQL extra config file. In order to do so, please execute `make tests-prep`.
-
-The MySQL server configuration must include `my-ssl.cnf` file,
-which can be found in `test/ssl/`.
-**Do not run** `make tests-prep` after you start MySQL,
-because CA certificates will no longer match.
+To quickly setup a test environment, execute `.ci/prep.sh`, then execute `make tests`.
```
zuiderkwast, line 120: If the `.ci/` stuff is used also for local tests, then maybe we should name it something other than CI? And not a hidden directory? Something under `test` such as `test/setup/` or `test/util/` or something like that, or is it not good to put things under test? (We already have `test/ssl/`. (If you don't like to put scripts under test/, then we can move the ssl cert scripts too, to keep all in one place.)

Do you think nobody should run the tests without docker from now on? I'm not sure we should require running a script that needs sudo.

The scripts are already divided into the ones that are used for docker and the ones that are used also when running without docker. For example `.ci/init.sh` is the script that creates the SQL users and `make test-prep` sets up the certificates. We can delete the SQL code but maybe keep some of the explanations about which steps are required and what each of the scripts do.

zuiderkwast:
> > > I don't know what's the problem with the test case for decimals that was commented out. I don't remember any problem with it in the past. It was a long time ago.
> > 
> > 
> > @zuiderkwast thank you for the review. I need a pointer to see how this test worked before, and compare the test runs to see why it's not working now. I tried to test with MySQL 8.4, 8.0 and 5.7, none of them worked.
> 
> Actually, re-ran it against 5.7, it worked there. On 8.0:
> 
> ```
> **error:{badmatch,{ok,ok,
>               [{warning_msg,"Note 1265: Data truncated for column 'balance' at row 1\n in UPDATE test_decimals SET balance = balance - ? WHERE id = ?\n"}]}}
>   output:<<"">>
> ```

Interesting. There is already some handling for this warning, for the other cases where we try to send the data as float instead of decimal, like here: https://github.com/mysql-otp/mysql-otp/blob/master/test/mysql_tests.erl#L815

I haven't looked into how to send the decimal type properly, but the idea that it shouldn't be sent as a float, to prevent rounding errors. It was contributed by (someone, don't remember who).

zmstone:
> I haven't looked into how to send the decimal type properly, but the idea that it shouldn't be sent as a float, to prevent rounding errors. It was contributed by (someone, don't remember who).

It's encoded as `0x00, VarLen, FloatNumberStringFormat`

https://github.com/mysql-otp/mysql-otp/blob/011702f7f06ff7bdb5530ef4ac6eb6417e1e913c/src/mysql_protocol.erl#L1127-L1134

The commit which introduced this test:
https://github.com/mysql-otp/mysql-otp/commit/a74aff1e45b5df26240bcf2e30fe2e516cf4e2a0

README.md
zmstone: > If the .ci/ stuff is used also for local tests, then maybe we should name it something other than CI? 

Yes, good point. I can do this in a follow up PR.
The priority of this PR is to get GitHub Actions up and running to make future PRs verified.

> Do you think nobody should run the tests without docker from now on? I'm not sure we should require running a script that needs sudo.

Everyone wants a quick feedback loop should use the scripts provided in this repo to setup MySQL.
The unix domain socket tests will not work pass otherwise.

zuiderkwast:
I think the test case looks correct. I also checked how expressions like `balance = balance - ?` should be handled here: https://dev.mysql.com/doc/refman/8.4/en/precision-math-expressions.html. It looks like it should be handled without truncation if all values in the expression are exact (decimal or integer).

So, can we assume that MySQL 8 has some bug here? We can conditionally run the test case only on 5.7 where it's known to work, and maybe on MariaDB if it works there (i didn't check). Or accept a warning for 8.x and no warning for 5.7. I would be fine with that.

zmstone:
> So, can we assume that MySQL 8 has some bug here? We can conditionally run the test case only on 5.7 where it's known to work, and maybe on MariaDB if it works there (i didn't check). Or accept a warning for 8.x and no warning for 5.7. I would be fine with that.

Just to add: I tested MySQL 8.0, 8.4 and 9.0 all emitted the same warning. 
I have not tried MriaDB yet, will find time to do so.

README.md
zuiderkwast: I think it can be refactored a bit to separate what needs sudo, what runs with docker, what doesn't.

As of now, I don't want to delete the README instructions about setting up running the tests locally without docker though, so please undelete those parts, but we can keep the link to the `.ci` script too.

test/mysql_tests.erl
```diff
@@ -824,18 +830,38 @@ decimal_trunc(_Pid) ->
...
+assert_decimal_trunctation_warning([]) -> ok;
+assert_decimal_trunctation_warning([{warning_msg, Msg}]) ->
+    ?assertMatch("Note 1265: Data truncated for column 'balance'" ++ _, Msg),
+    case os:getenv("MYSQL_IMAGE") of
+        "mariadb" ->
+            throw("not expecting mariadb to emit warning " ++ Msg);
+        _ ->
+            ok
+    end,
+    case {os:getenv("MYSQL_IMAGE"), os:getenv("MYSQL_VERSION")} of
```
zuiderkwast, line 858: Let's avoid depending on external variables. There are already mechanisms in this test suite to ask the server for its version: `db_version_string(Pid)`, `parse_db_version(Version)` and `is_mariadb(Version)`.

Look at the conditional code for the JSON test for inspiration:

```erlang
json(Pid) ->
    Version = db_version_string(Pid),
    try
        is_mariadb(Version) andalso throw(no_mariadb),
        Version1 = parse_db_version(Version),
        Version1 >= [5, 7, 8] orelse throw(version_too_small)
    of _ ->
        test_valid_json(Pid),
        test_invalid_json(Pid)
    catch
        throw:no_mariadb ->
            error_logger:info_msg("Skipping JSON test, not supported on"
                                  " MariaDB.~n");
        throw:version_too_small ->
            error_logger:info_msg("Skipping JSON test. Current MySQL version"
                                  " is ~s. Required version is >= 5.7.8.~n",
                                  [Version])
    end.
```

README.md
zmstone: In the latest commit, I added instructions about how to run tests against MySQL or MariaDB running in localhost.

I am not convinced we should keep the SQL statements in README because it's hard to keep it up-to-date.
It should be enough to reference to init.sh from README

zmstone:
image

zmstone:
Hi @zuiderkwast thank you for the review. 
Please let me know if there is anything else to be fixed.

zuiderkwast (approved):
Yeah, I think this is good to merge. Leaving it open for a few more days though.

@zmstone Can you update the PR description to describe the change with a little more details, including the fixes for decimal tests and the +0.0 Erlang warning thing? Then I'll squash-merge and use this text as the commit message. I want good commit messages because they're used for the changelog too.

@juhlig Please take a look if you have time.

zmstone:
> @zmstone Can you update the PR description to describe the change with a little more details, including the fixes for decimal tests and the +0.0 Erlang warning thing? 

done.

test/mysql_tests.erl
```diff
@@ -264,76 +277,59 @@ reset_connection_test() ->
...
+            mysql:stop(Pid2);
+        {error, eafnosupport} ->
+            error_logger:info_msg("Skipping unix socket test. "
+                                    "Not supported on this OS.~n")
+    end.
+
+socket_backend_test() ->
+    case mysql:start_link([{user, ?user},
+                            {password, ?password},
+                            {tcp_options, [{inet_backend, socket}]}])
```
juhlig, line 314: Indentation.
```suggestion
    case mysql:start_link([{user, ?user},
                           {password, ?password},
                           {tcp_options, [{inet_backend, socket}]}])
```

test/mysql_tests.erl
```diff
@@ -264,76 +277,59 @@ reset_connection_test() ->
...
+            CiDir
+    end.
+
 unix_socket_test() ->
-    try
-        list_to_integer(erlang:system_info(otp_release))
+    %% Get socket file to use
+    Dir = run_dir(),
+    {ok, Pid1} = mysql:start_link([{user, ?user},
+                                    {password, ?password}]),
```
juhlig, line 295: Indentation.
```suggestion
    {ok, Pid1} = mysql:start_link([{user, ?user},
                                   {password, ?password}]),
```

test/mysql_tests.erl
```diff
@@ -264,76 +277,59 @@ reset_connection_test() ->
...
+    {ok, [<<"@@socket">>], [[SockFile0]]} = mysql:query(Pid1, "SELECT @@socket"),
+    SockFile = filename:join([Dir, filename:basename(SockFile0)]),
+    mysql:stop(Pid1),
+    %% Connect through unix socket
+    case mysql:start_link([{host, {local, SockFile}},
+                            {user, ?user}, {password, ?password}]) of
```
juhlig, line 301: Indentation.
```suggestion
    case mysql:start_link([{host, {local, SockFile}},
                           {user, ?user}, {password, ?password}]) of
```

test/mysql_tests.erl
```diff
@@ -264,76 +277,59 @@ reset_connection_test() ->
...
+        {ok, Pid2} ->
+            ?assertEqual({ok, [<<"1">>], [[1]]},
+                            mysql:query(Pid2, <<"SELECT 1">>)),
+            mysql:stop(Pid2);
+        {error, eafnosupport} ->
+            error_logger:info_msg("Skipping unix socket test. "
+                                    "Not supported on this OS.~n")
```
juhlig, line 308: Indentation.
```suggestion
            error_logger:info_msg("Skipping unix socket test. "
                                  "Not supported on this OS.~n")
```

test/mysql_tests.erl
```diff
@@ -264,76 +277,59 @@ reset_connection_test() ->
...
     ok.
 
+run_dir() ->
+    CiDir = filename:join(["scripts", "run"]),
+    case filelib:wildcard("*.pid", CiDir) of
+        [] ->
+            %% Assume MySQL or MariaDB is running in host
+            "/var/run/mysqld/";
+        _ ->
+            %% This is the mouted dir for MySQL or MariaDb running in docker
```
juhlig, line 287: Typo.
```suggestion
            %% This is the mounted dir for MySQL or MariaDb running in docker
```

test/mysql_tests.erl
```diff
@@ -833,9 +823,25 @@ decimal_trunc(_Pid) ->
...
+        true ->
+            throw("Not expecting mariadb to emit warning: " ++ Msg);
+        false ->
+            case parse_db_version(VersionStr) of
+                [5 | _] ->
+                    throw("Not expecting MySQL 5.x to emit warning: " ++ Msg);
+                _ ->
+                    ok
+            end
+    end.
```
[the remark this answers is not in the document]
zuiderkwast: @juhlig Do you have any clue why MySQL does emit a warning for this test case? I'm wondering if there is any appropriate way at all to use the decimal type without a warning, from any client.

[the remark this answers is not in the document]
juhlig: Sorry, I have no idea 🤷‍♂️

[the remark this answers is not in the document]
zmstone: @juhlig  thanks for the review. changed it to be more defensive

juhlig:
I'll have a second look on monday, but looks pretty much ok to me now 🙂

juhlig (approved):
Sorry for the delay, I was very busy the last couple of days =^^=

README.md
```diff
@@ -29,7 +28,7 @@ Some of the features:
 
 Requirements:
 
-* Erlang/OTP version R16B or later
+* Erlang/OTP version 19 or later
```
juhlig, line 31: @zuiderkwast it surprises me a little that you are willing to go this step since you always seemed to take special pride in being compatible all the way back to R16 :sweat_smile: But I'm really in favor of this step. In fact, it would allow us to remove the backporting of `gen_server:close` which we have somewhere in there, since that appeared in OTP 18.

zuiderkwast: Haha, yeah, after many years we can make a small bump.

We could probably go as fast as OTP 22 or something, but it annoys me that tools like rebar3 only support the two latest major versions. That's really to little. We should at least cover what's included in the oldest LTS distros that are still not yet end-of-life. It used to be CentOS but now I guess it's AlmaLinux or something.

zuiderkwast: It's `gen_server:stop`, right?

zuiderkwast:
Merged! Build passed! Thanks! :100:

Looking forward to follow-ups:
* delete the Tavis CI file.
* remove the backporting of `gen_server:stop` which we have somewhere in there, since that appeared in OTP 18.

README.md
juhlig: > It's `gen_server:stop`, right?

Yeah, that's what I meant :)

PR #209, opened 2024-10-28
Add rebar.config and change default build tool to rebar3

zmstone:
Summary:

- Added rebar.config, and changed CI builds to use rebar3
- Fixed edoc issue found in rebar3 test
- Fixed dialyzer issue found in rebar3 test

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,10 +28,16 @@ jobs:
         with:
           otp-version: ${{matrix.otp}}
           rebar3-version: 3.20.0
-      - name: Set up environment variables
+      - name: Run tests and doc generation
         run: |
           set -x
           export MYSQL_IMAGE=${{ matrix.image[0] }}
           export MYSQL_VERSION=${{ matrix.image[1] }}
+          rebar3 xref
+          rebar3 dialyzer
           make tests-prep
           make tests
+          # also run rebar3 eunit
+          rebar3 eunit -v --cover
+          rebar3 cover
+          rebar3 edoc
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,5 @@ _build/
 scripts/run/
 scripts/certs/
 scripts/my-ssl.cnf
+mysql_dialyzer_*
+*.tmp
--- a/rebar.config
+++ b/rebar.config
@@ -1,0 +1,27 @@
+{edoc_opts, [
+    {doclet, edoc_doclet},
+    {stylesheet, "priv/edoc-style.css"},
+    {todo, true},
+    {overview, "doc/overview.edoc"},
+    {subpackages, true}
+]}.
+
+{dialyzer, [
+    {warnings, [unmatched_returns, error_handling]},
+    {plt_location, "."},
+    {plt_prefix, "mysql_dialyzer"},
+    {plt_apps, all_apps},
+    {statistics, true}
+]}.
+
+{cover_opts, [verbose]}.
+{cover_export_enabled, true}.
+
+{xref_checks, [
+    undefined_function_calls,
+    undefined_functions,
+    locals_not_used,
+    deprecated_function_calls,
+    warnings_as_errors,
+    deprecated_functions
+]}.
--- a/src/mysql.app.src
+++ b/src/mysql.app.src
@@ -21,5 +21,5 @@
     {vsn, "1.8.0"},
     {modules, []},
     {registered, []},
-    {applications, [kernel,stdlib,ssl]}
+    {applications, [kernel,stdlib,ssl,public_key]}
 ]}.
--- a/src/mysql.erl
+++ b/src/mysql.erl
@@ -490,7 +490,7 @@ execute(Conn, StatementRef, Params, FilterMap) when FilterMap == no_filtermap_fu
 %% See `query/5' for an explanation of the `FilterMap' argument.
 %%
 %% Note that if this function is called on a connection which is already in transaction 
-%% owned by another process, `{error, busy}` will be returned.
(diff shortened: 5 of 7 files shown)


pankajmyt:
What's the rationale behind this?

zmstone:
> What's the rationale behind this?

Because rebar3 is the de facto standard in the Erlang ecosystem, backed by the Erlang/OTP team and widely supported in the community.

zuiderkwast:
AFAIK, the OTP team is not using nor maintaining rebar3. It just lives under the erlang umbrella. Erlang/OTP itself is built using make files, and AFAIK other old Erlang projects within Ericsson that predate rebar3 don't use it much either. Others use different tools too, like WhatsApp are using some other build tool, IIUC.

I didn't like rebar3 much initially, but I guess it's much better now after nearly 10 years. I didn't like it partly because of the difficulty to see exactly what it's doing, i.e. abstracted complexity. I guess I'm part of a small minority in this regard. :grin: To be fair, I don't like erlang.mk either because of its complexity. In this project, we don't have any dependencies so it's very easy to build using any tool. The aim for this project was always to let the users built it with any tool that can build OTP [applications](https://www.erlang.org/doc/system/applications). It's already possible to rebar3 for example. It doesn't need a rebar.conf.

Potentially breaking change: The current behaviour is that if you build with make, the beam files end up under `ebin/` but if you build with rebar3, they end up somewhere else. This may break some setups. Nothing impossible to fix for users, but it can be annoying, besides building with make now requires rebar3 to be installed too.

So far, while rebar3 can be used when just building it, we've been using the Makefile for development purposes, including building github pages and some other special stuff. I could live with rebar3 for this, but using both a makefile and rebar config seems a bit too much. If we migrate to rebar3, I'd prefer that we move everything from the makefile to rebar3 scripts or configs. I'm not sure rebar3 is good for various custom non-erlang stuff though. Is it? Apparently, we even use mix rather than rebar3 for publishing to hex.

I generally like make because it's explicit and declarative and can build anything. Regardless what we decide, we could drop erlang.mk. Compiling erl to beam using erlc is trivial and OTP tools like eunit and dialyzer are very easy to run using a few lines. My favourite would be a small makefile with simple rules like

```make
compile: $(beams)

erls = $(wildcard src/*.erl)
beams = $(erls:src/%.erl=ebin/%.beam)

ebin/%.beam: src/%.erl | ebin
	erlc $(ERLC_OPTS) -o ebin $<

ebin:
	mkdir -p ebin
```

zmstone:
I liked erlang.mk too before rebar3 (or maybe during the early versions of rebar3).
I still like Makefile instead of using rebar3 for everything.
Rebar3 supports publishing to hex, there is no good reason to replace the mix command in this project, after all this can be a dependency for Elixir projects too.

pankajmyt:
Used to work extensively with erlang, but last few years moved to golang. So much thanks for the input. I like the simplicity of erlang.mk build setup.

zmstone:
@zuiderkwast @juhlig wdyt?

juhlig:
No objections, but isn't it possible to support _both_ rebar3 _and_ erlang.mk? At least I think that was possible in the past.

zmstone:
> No objections, but isn't it possible to support _both_ rebar3 _and_ erlang.mk? At least I think that was possible in the past.

yes, it's possible. will bring it back then.

zmstone:
> > No objections, but isn't it possible to support _both_ rebar3 _and_ erlang.mk? At least I think that was possible in the past.
> 
> yes, it's possible. will bring it back then.

Sorry, second thoughts: I don’t really see why should we support both, this app is used as dependency, other erlang.mk projects can work with rebar projects just fine.

zuiderkwast:
The difference is that we start requiring rebar when building with make. I don't think e.g. cowboy does that.

zmstone:
> The difference is that we start requiring rebar when building with make. I don't think e.g. cowboy does that.

Done adding back erlang.mk

zmstone:
gentle ping

zuiderkwast (commented):
Sorry, I don't see much value in this change. I feel I need to find good answers to all the changes and I don't have that much time for that right now.

juhlig:
> Sorry, I don't see much value in this change. I feel I need to find good answers to all the changes and I don't have that much time for that right now.

I was somewhat indifferent regarding the issue of the (primary) build tool, with a tendency to stick to `erlang.mk`. (Personally, I started out with and still prefer `erlang.mk`, never really got to grips with `rebar` :man_shrugging:)
That said, I also see little need or value in this change, like what things that we want or need can we do with `rebar` that we can't do with `erlang.mk`?

That is not to say that we won't _ever_ want or need to change. I suggest keeping this PR around for a while, see what comes up. But generally, ATM I don't think we should exchange one thing that has served us well in the past and never (?) given raise to complaints for another just because we can.

> The difference is that we start requiring rebar when building with make. I don't think e.g. cowboy does that.

`cowboy` is probably not the best example for this case, since `erlang.mk` was made and maintained by the same guy who also made and maintains `cowboy` and related projects (`ranch`, `gun`, `cowlib`, ...), and it probably was made for just that purpose in the first place :wink:

That said, IIRC `erlang.mk` pulls in and uses `rebar` when it needs to, eg in order to compile a dependency. Not sure `rebar` does vice versa. "Erlang in Anger"  says that `erlang.mk` dependencies "tend to" be supported, not sure if this is still true and how far it goes :man_shrugging:

zmstone:
Restored everything, but just added `rebar.config` and updated CI to use rebar3 commands.
Also updated PR summary.

zmstone:
If you want edoc, dilayzer and cover fixes in the CI, be my guest close this PR, my time is worth spent doing something else.

PR #211, opened 2024-11-05
Remove exception handling macros

juhlig:
Since we are now targeting OTP>=22, we can also remove the exception handling switch macros, as `Class:Reason:Stacktrace` is supported since OTP 21.

### DIFF
--- a/include/exception.hrl
+++ b/include/exception.hrl
@@ -1,7 +1,0 @@
--ifdef(OTP_RELEASE). %% this implies 21 or higher
--define(EXCEPTION(Class, Reason, Stacktrace), Class:Reason:Stacktrace).
--define(GET_STACK(Stacktrace), Stacktrace).
--else.
--define(EXCEPTION(Class, Reason, _), Class:Reason).
--define(GET_STACK(_), erlang:get_stacktrace()).
--endif.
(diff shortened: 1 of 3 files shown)


juhlig:
Wait, what was I thinking... we are targeting OTP 19+ now, not 22+ :sweat_smile:

zuiderkwast:
We can be targeting >= 22 if you want. :)

zuiderkwast:
> Eerie... who are you, and what did you do with the _real_ zuiderkwast? 😁

I decided that it's better to be practical. :laughing: We don't need to support releases that don't even receive security patches for years, but I don't want to drop releases that may still in use in LTS distros. Debian Stable currently provides OTP 25 (see https://packages.debian.org/stable/erlang). Ubuntu Focal 2020 is still supported and it provides OTP 22. Ubuntu provides some extended support for even older releases; see https://ubuntu.com/about/release-cycle. I believe RHEL do the same.

Currently, we can draw the line somewhere around OTP 21 or 22.

zuiderkwast (commented):
Can you update the README too?

juhlig:
Done :)

zuiderkwast (approved):

PR #212, opened 2024-11-06
Hide sensitive information in state and use `gen_server:format_status/1,2` depending on OTP release

juhlig:
We use the `gen_server:format_status/2` callback in order to hide sensitive data from logs.

In OTP/25, another callback `format_status/1` was introduced which should replace `format_status/2`. In OTP/27, `format_status/2` was deprecated, and raises a compiler warning.

This PR adds the `format_status/1` callback, and a conditional that adds the `format_status/2` callback for OTP<25, where `format_status/1` is absent.

I also changed the behavior such that it _always_ hides the sensitive data, ie not only in case of crashes, because BSTS ;) If a user _needs_ the unobfuscated state, he can always use `sys:get_state`.

### DIFF
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -28,7 +28,11 @@
 
 -behaviour(gen_server).
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2,
-         code_change/3, format_status/2]).
+         code_change/3, format_status/1]).
+
+-if(?OTP_RELEASE < 25).
+-export([format_status/2]).
+-endif.
 
 -define(default_host, "localhost").
 -define(default_port, 3306).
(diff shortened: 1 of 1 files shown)


src/mysql_conn.erl
```diff
@@ -851,18 +855,25 @@ demonitor_processes([{_FromPid, MRef}|T], Count) ->
...
-        ),
-	{data, [{"State", State#state{password = hidden, ssl_opts=SSLOpts1}}]}.
+format_status(Status = #{state := State=#state{ssl_opts=SSLOpts}}) ->
+    Status#{state := State#state{password = hidden,
+				 ssl_opts=obfuscate_ssl_options(SSLOpts)}}.
+
+-if(?OTP_RELEASE < 25).
+format_status(_, [_PDict, State=#state{ssl_opts=SSLOpts}]) ->
+    {data, [{"State", State#state{password = hidden,
+				  ssl_opts=obfuscate_ssl_options(SSLOpts)}}]}.
```
zuiderkwast, line 865: I'm not sure it's consistent in this repo, but I think the idea is to always to have spaces around `=`.

```suggestion
format_status(Status = #{state := State = #state{ssl_opts = SSLOpts}}) ->
    Status#{state := State#state{password = hidden,
				 ssl_opts = obfuscate_ssl_options(SSLOpts)}}.

-if(?OTP_RELEASE < 25).
format_status(_, [_PDict, State = #state{ssl_opts = SSLOpts}]) ->
    {data, [{"State", State#state{password = hidden,
				  ssl_opts = obfuscate_ssl_options(SSLOpts)}}]}.
```

zuiderkwast (approved):

src/mysql_conn.erl
juhlig: Applied and squashed :)

zuiderkwast:
Brw, there's no need to squash and force-push. We squash-merge the PRs anyways when we merge. I like to use the PR title and description as the commit messages. (That's why I edit them somethimes before merging.)

PR #214, opened 2024-11-06
Fix `public_key:encrypt_public` deprecation warning

### LINKED ISSUE #213: `public_key:encrypt_public` (temporarily) deprecated
We use `public_key:encrypt_public` in `mysql_protocol`. This function was deprecated in OTP 27.0 and de-deprecated again in OTP 27.1. The reason for the deprecation was a vulnerability in RSA. The reason for the de-deprecation was that this vulnerability only affects `rsa_pkcs1_padding`, while `rsa_pkcs1_oaep_padding` seems to be fine.

Which of the two we use depends on the MySQL version: before 8.0.5, `rsa_pkcs1_padding`, since 8.0.5, `rsa_pkcs1_oaep_padding`. So this is pretty much out of our hands. That said, _both_ cause the deprecation warning in OTP 27.0, since it is the _function_ that was deprecated, and with `warnings_as_errors` this prevents compilation.

So we are faced with two issues here, which only affects tiny windows of versions. The deprecation warning issue only affects OTP 27.0, The security issue affects only MySQL 8.0.0 up to 8.0.5, and is out of our hands.

While the affected windows are tiny, I still think we should do something about it. I found a trick to mitigate the deprecation warning, that is easily taken care of. As for the security issue, I think all we can (and should) do is log a warning in such cases?

WDYT?

juhlig:
Fixes #213. (Suggestion)

### DIFF
--- a/src/mysql_protocol.erl
+++ b/src/mysql_protocol.erl
@@ -168,7 +168,7 @@ auth_finish_or_switch(AuthPluginName, AuthPluginData, Password,
             auth_finish_or_switch(AuthPluginName, AuthPluginData, Password, SockModule,
                                   Socket, ServerVersion, SeqNum2);
         {public_key, PubKey} ->
-            %% Serveri has sent its public key (certainly specific to the caching_sha2_password
+            %% Server has sent its public key (certainly specific to the caching_sha2_password
             %% method). We encrypt the password with the public key we received and send
             %% it back to the server.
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
Great, sorry for the delay.

Can you copy the relevant description from the issue to this PR's description? Then we'll use this as the commit message when squash-merging.

juhlig:
I reworded the description to what I think is usable in a commit message, so you just have to drop the "Fixes #213." part when merging.

PR #215, opened 2024-11-20
Call SockMod:close before stop

zmstone:
When a transaction caller process exits, `mysql_conn` will close the socket and stop itself.
Prior to this fix, the socket close function call disregards the transport implementation module (`sockmod`), always calls `gen_tcp:close` and matches it against `ok`. This PR changes to `ok = SockMod:close(...),`.

Additionally, the `flush` option is added to the `demonitor` call, when a transaction monitor is demonitored.

----

I'd like to discuss though if it really makes sense to stop with a reason other than `shutdown` or `normal`,
because this will cause a crash log emitted regardless of the reason of owner processes' exit reason. 

Below is a crash log example:

```
.=ERROR REPORT==== 20-Nov-2024::14:18:33.240184 ===
Connection Id 15 closing with reason: {application_process_died,<0.453.0>}

=ERROR REPORT==== 20-Nov-2024::14:18:33.240526 ===
** Generic server tardis terminating
** Last message in was {'DOWN',#Ref<0.1068720556.291766273.97155>,process,
                               <0.453.0>,normal}
** When Server state == {state,
                            [8,4,2],
                            undefined,undefined,ssl,[],
                            [{server_name_indication,disable},
                             {cacertfile,"test/ssl/ca.pem"}],
                            "localhost",3306,"otptestssl",hidden,undefined,
                            ["SET @foo = 'bar'","SELECT 1",
                             "SELECT 1; SELECT 2"],
                            [{foo,"SELECT @foo"}],
                            <<"caching_sha2_password">>,
                            <<125,108,113,76,78,114,60,14,112,9,20,96,120,33,
                              122,106,63,94,124,8,0>>,
                            [],true,false,5000,infinity,infinity,60000,0,3,0,
                            0,
                            [{<0.453.0>,#Ref<0.1068720556.291766273.97155>}],
                            undefined,
                            {dict,1,16,16,8,80,48,
                                {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                                 []},
                                {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                                  [],
                                  [[foo|{prepared,1,"SELECT @foo",0,0}]]}}},
                            empty,false,false,auto}
** Reason for termination ==
** {application_process_died,<0.453.0>}
```

### DIFF
--- a/test/ssl_tests.erl
+++ b/test/ssl_tests.erl
@@ -27,7 +27,7 @@
 -define(cacertfile,   "test/ssl/ca.pem").
 
 successful_ssl_connect_test() ->
-    [ application:start(App) || App <- [crypto, asn1, public_key, ssl] ],
+    ok = start_apps(),
     common_basic_check([{ssl, [{server_name_indication, disable},
                                {cacertfile, ?cacertfile}]},
                         {user, ?ssl_user}, {password, ?ssl_password}]),
@@ -48,3 +48,25 @@ common_basic_check(ExtraOpts) ->
 common_conn_close() ->
     Pid = whereis(tardis),
     mysql:stop(Pid, 5000).
+
+owner_down_test() ->
+    ok = start_apps(),
+    Options = [{name, {local, tardis}},
+               {queries, ["SET @foo = 'bar'", "SELECT 1",
+                          "SELECT 1; SELECT 2"]},
+               {prepare, [{foo, "SELECT @foo"}]},
+               {ssl, [{server_name_indication, disable},
+                      {cacertfile, ?cacertfile}]},
+               {user, ?ssl_user}, {password, ?ssl_password}],
+    {ok, Pid} = mysql:start_link(Options),
+    unlink(Pid),
+    Mref = erlang:monitor(process, Pid),
+    Caller = spawn(fun() -> gen_server:call(Pid, start_transaction, 10) end),
+    receive
+        {'DOWN', Mref, _, Pid, Reason} ->
+            ?assertEqual({application_process_died, Caller}, Reason)
+    end.
+
+start_apps() ->
+    {ok, _} = application:ensure_all_started(ssl),
+    ok.
--- a/test/ssl_tests.erl
+++ b/test/ssl_tests.erl
@@ -65,6 +65,8 @@ owner_down_test() ->
     receive
         {'DOWN', Mref, _, Pid, Reason} ->
             ?assertEqual({application_process_died, Caller}, Reason)
+    after 1000 ->
+        error(down_signal_not_received)
     end.
 
 start_apps() ->
--- a/src/mysql_conn.erl
+++ b/src/mysql_conn.erl
@@ -497,7 +497,7 @@ handle_call(rollback, {FromPid, _},
                            decode_decimal = DecodeDecimal,
                            transaction_levels = [{FromPid, MRef} | L]})
   when Status band ?SERVER_STATUS_IN_TRANS /= 0 ->
-    erlang:demonitor(MRef),
+    erlang:demonitor(MRef, [flush]),
     Query = case L of
         [] -> <<"ROLLBACK">>;
         _  -> <<"ROLLBACK TO s", (integer_to_binary(length(L)))/binary>>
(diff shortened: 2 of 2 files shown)


zuiderkwast (approved):
Thanks! Implementation looks good. This must have been missed when TLS was added. Demonitor with flush is also good.

Regarding the exit reason, I'm not sure. If the calling process exits in the middle of a transaction, isn't that always an error case? Can you think of a use case where this is not an error?

If the calling process in the transaction is not the owner of the client process (owner = the process that called start_link) then I think it's unexpected that we exit with reason 'ok'.

Reason 'shutdown' is normally only caused by a supervisor. It does `exit(Pid, shutdown)` and then the process exits with the same reason.

zmstone:
> Thanks! Implementation looks good. This must have been missed when TLS was added. Demonitor with flush is also good.
> 
> Regarding the exit reason, I'm not sure. If the calling process exits in the middle of a transaction, isn't that always an error case? Can you think of a use case where this is not an error?
> 
> If the calling process in the transaction is not the owner of the client process (owner = the process that called start_link) then I think it's unexpected that we exit with reason 'ok'.
> 
> Reason 'shutdown' is normally only caused by a supervisor. It does `exit(Pid, shutdown)` and then the process exits with the same reason.

There are 2 cases
1. The application is on the way to shutdown
2. The the caller had a bug

For casa 1, I may argue at least for `normal`, `shutdown` and `{shutdown, _}`, there should not be a crash log
For case 2, it's a judgement call, IMO, it's the caller process' supervisor or linked owner to log the error but not the other way around. Fact is, people mostly do it already and end up having too much logs.

`{shutdown, Reason}` is a common practice to avoid crash logs while still for the monitors to get `Reason` instead of just `shutdown`.

zmstone:
don't want to stall the process of this PR.
Feel free to merge, I can send another PR to change it to a `{shutdown, ...}` terminate reason if you think it's a good idea.

zuiderkwast:
> don't want to stall the process of this PR. Feel free to merge, I can send another PR to change it to a `{shutdown, ...}` terminate reason if you think it's a good idea.

Yes, let's discuss this in a separate PR. I think there is something more to discuss here.

For example, if the client is started by a supervisor with restart = transient, then it is not restarted if it exits with reason `ok`, `shutdown` or `{shutdown, _}`. Is that a problem in any case? Maybe not. I don't know yet, or if there is some other case like that. :)

zuiderkwast:
@juhlig Do you want to take a look before merging?

juhlig:
I think this PR is fine. Good catch re `gen_tcp` vs `SockMod` :+1: Including `flush` in `demonitor` is also a good thing. So from my side, approved :slightly_smiling_face: 

I don't think changing the exit reason to always be normal-ish is a good idea though, at least I have to think about this a bit more. OTOH, it prevents "spamming the logs", but OTOH it brushes something that went wrong under the carpet and makes debugging harder. What I mean is, you can always _ignore_ surplus logs if you don't care, but you can't un-ignore stuff that is not there if you _do_ care.

tl;dr: I'm all for merging this PR as it stands :+1: For the exit reason thing, which is to be done in a separate PR, I'll think about it a bit more ;)

test/ssl_tests.erl
```diff
@@ -48,3 +48,25 @@ common_basic_check(ExtraOpts) ->
...
+                      {cacertfile, ?cacertfile}]},
+               {user, ?ssl_user}, {password, ?ssl_password}],
+    {ok, Pid} = mysql:start_link(Options),
+    unlink(Pid),
+    Mref = erlang:monitor(process, Pid),
+    Caller = spawn(fun() -> gen_server:call(Pid, start_transaction, 10) end),
+    receive
+        {'DOWN', Mref, _, Pid, Reason} ->
+            ?assertEqual({application_process_died, Caller}, Reason)
+    end.
```
juhlig, line 68:
```suggestion
    receive
        {'DOWN', Mref, _, Pid, Reason} ->
            ?assertEqual({application_process_died, Caller}, Reason)
    after 1000 ->
        error(down_signal_not_received)
    end.
```

juhlig: On second thought, it would be good to have a timeout in the test.

PR #216, opened 2024-11-29
Replace `error_logger` with logger

juhlig:


test/logger_acc.erl
```diff
@@ -0,0 +1,98 @@
...
+%% @doc Executes `Fun' and captures all logged errors returns as well as
+%% uncaught errors in `Fun'.
+-spec capture(fun (() -> ResultOfFun)) ->
+    {ok, ResultOfFun, AccumulatedErrors} |
+    {throw | error | exit, Reason, Trace, AccumulatedErrors}
+  when ResultOfFun :: term(),
+       Reason :: term(),
+       Trace :: list(),
+       AccumulatedErrors :: [{error|warning_msg|info_msg, string()} |
+                             {error_report|warning_report|info_report, term()}].
```
zuiderkwast, line 19: These need to be updated? From the usage of this, I see it's something like this:

    {error | warning | info, [atom()], string()}

test/logger_acc.erl
```diff
@@ -0,0 +1,98 @@
...
+flush_logs(AccPid, Tag) ->
+    AccMon = monitor(process, AccPid),
+    AccPid ! {Tag, flush},
+    receive
+        {Tag, Events} ->
+            demonitor(AccMon, [flush]),
+            Events;
+        {'DOWN', AccMon, process, AccPid, Reason} ->
+            error({accumulator_process, Reason})
+    end.
```
zuiderkwast, line 48: Do we need a timeout for this receive? Or not needed when we have the monitor?

test/logger_acc.erl
```diff
@@ -0,0 +1,98 @@
...
+
+log_acc_loop(Tag, Parent) ->
+    log_acc_loop(Tag, Parent, open, []).
+
+log_acc_loop(Tag, Parent, Status, Acc) ->
+    Timeout = case Status of
+        open -> infinity;
+	closing -> 1000
+    end,
```
zuiderkwast, line 57: When I wrote that code 10-ish years ago, I didn't know the proper indentation of Erlang code. :)

Are those tabs btw? Please use spaces.

```suggestion
    Timeout = case Status of
                  open -> infinity;
                  closing -> 1000
              end,
```

zuiderkwast (commented):
Nice! Thanks!

test/logger_acc.erl
juhlig: I went for `{logger:level(), [atom()], string()|logger:report()}`.

test/logger_acc.erl
juhlig: No need I think, from how the accumulating loop is designed.

test/logger_acc.erl
juhlig: Indeed... they keep sneaking in :sweat_smile:

juhlig: > When I wrote that code 10-ish years ago, I didn't know the proper indentation of Erlang code. :)

I'm going back and forth between different styles, and I see that I even used 2 different ones in this very function :sweat_smile: I think the style used by OTP itself is to align the `case` and `end` in the same column, and to indent with 4 spaces starting from that column.

juhlig: All fixed in last commit.

juhlig:
All fixed in last commit.

test/logger_acc.erl
zuiderkwast: Yes. I like this style, and it's what you get if you're using emacs. (I do, nowdays. It's also possible to run it in CI with erlfmt.)

test/logger_acc.erl
zuiderkwast: OK, and it's just a testing helper, so worst case the test will time out.

test/logger_acc.erl
```diff
@@ -0,0 +1,97 @@
...
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+
+capture_success_test() ->
+    Result = ?MODULE:capture(fun () ->
+        logger:notice("Hello ~p", [world]),
+        logger:notice("Hello ~p", [again]),
+        foo
+    end),
```
zuiderkwast, line 87: This indentation too should probably be `end` under `fun`, but I will not insist. Let's merge it. :)

zuiderkwast (approved):

test/logger_acc.erl
juhlig: I'm sticking to vim :blush:

test/logger_acc.erl
juhlig: That's quickly done, see the last commit :)

zuiderkwast: Thx!

OTP style (emacs) indents the body of a fun 8 spaces thouh, because a fun can have multiple clauses.

```
fun (foo) ->
        bar;
    (baz) ->
        quux
end
```

PR #218, opened 2024-11-29
Update `erlang.mk`

juhlig:
Updates our 7 years old version of `erlang.mk` :) The core module `core/beam-cache` is now also needed to run tests.

### DIFF
--- a/erlang.mk
+++ b/erlang.mk
@@ -17,16 +17,16 @@
 ERLANG_MK_FILENAME := $(realpath $(lastword $(MAKEFILE_LIST)))
 export ERLANG_MK_FILENAME
 
-ERLANG_MK_VERSION = 2017.08.28-5-g55699ee-dirty
+ERLANG_MK_VERSION = 2022.05.31-114-g509b15b-dirty
 ERLANG_MK_WITHOUT = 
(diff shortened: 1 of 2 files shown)


erlang.mk
```diff
@@ -47,18 +47,26 @@ verbose_0 = @
...
+
 # Temporary files directory.
 
 ERLANG_MK_TMP ?= $(CURDIR)/.erlang.mk
 export ERLANG_MK_TMP
 
 # "erl" command.
 
-ERL = erl +A0 -noinput -boot start_clean
+ERL = erl -noinput -boot no_dot_erlang -kernel start_distribution false +P 1024 +Q 1024
```
zuiderkwast, line 69: Does this work with the old OTP versions we claim to support?

juhlig: I tested `make`, `make tests`, `make docs` and `make dialyze` with OTP versions 21+, which is all we now support.

The reason for updating `erlang.mk` was that with the old version, `make dialyze` would not run on recent OTP versions because it was using a warning option that was removed at some point, namely `race_conditions`.

zuiderkwast (approved):

PR #219, opened 2024-12-02
Fix test for `socket` backend

juhlig:
The socket backend that `mysql_tests/socket_backend_test` tests is only supported in OTP 23+, and still very unstable in 23. Therefore, the change in this PR skips the test on OTP =< 23.

I also changed the level of logs emitted from the test suite from `info` to `notice`, because `info` is below the logging threshold in the default configuration, ie the minimum level is `notice`.

### DIFF
--- a/test/mysql_tests.erl
+++ b/test/mysql_tests.erl
@@ -316,8 +316,8 @@ socket_backend_test() ->
                           "Unstable on OTP 23.");
         _ ->
             {ok, Pid1} = mysql:start_link([{user, ?user},
-                                          {password, ?password},
-                                          {tcp_options, [{inet_backend, socket}]}]),
+                                           {password, ?password},
+                                           {tcp_options, [{inet_backend, socket}]}]),
             Dir = run_dir(),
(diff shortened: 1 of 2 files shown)


test/mysql_tests.erl
```diff
@@ -302,16 +302,22 @@ unix_socket_test() ->
...
+        Rel when Rel < 23 ->
+            logger:notice("Skipping socket backend test. "
+                          "Not supported on OTP ~b.", [Rel]);
+        23 ->
+            logger:notice("Skipping socket backend test. "
+                          "Unstable on OTP 23.");
+        _ ->
+            {ok, Pid1} = mysql:start_link([{user, ?user},
+                                          {password, ?password},
+                                          {tcp_options, [{inet_backend, socket}]}]),
```
zuiderkwast, line 320:
```suggestion
                                           {password, ?password},
                                           {tcp_options, [{inet_backend, socket}]}]),
```

zuiderkwast (approved):
OK, great.

Do we have any `logger:info` in the implementation code that should be changed to `notice` too?

juhlig:
> Do we have any `logger:info` in the implementation code that should be changed to `notice` too?

Nope, that was only in the tests.

